PR #492, opened 2015-01-15
Don't thrown NPE when item is not found in manifest (bad idref)

josteinaj:


rdeltour:
Rebased and cherry-picked as 412153520fa6e5da375db537661b785861601090

PR #497, opened 2015-02-14
Clarification of error message PKG_005.

skreutzer:
The extra fields feature of the ZIP format is forbidden for the mimetype file in the ZIP container because it would break support for magic numbers. The previous error message mislead to the impression that the error has something to do with filename extensions, while the extra field and the filename are separate things. Hopefully users will get enabled to search for ZIP extra fields and then look into the documentation of their ZIP generator to turn the extra fields feature off for the mimetype file or report an error for their EPUB generator program.


tofi86:
:+1:

This sounds reasonable and matches the actual checks performed. Going to merge this.

skreutzer:
I encountered it in an EPUB file, checked https://users.cs.jmu.edu/buchhofp/forensics/formats/pkzip.html against section 4: "ZIP Container" of the Open Container Format 2.0.1 v1.0.1 specification (page 18+) and can provide this screenshot of the effect:

extra_field

The actual mimetype "application/epub+zip" is supposed to start directly after mimetype, so reading systems don't need logic to skip over the extra field and can look for "application/epub+zip" from (including) byte 39.

tofi86:
Great, thanks for adding the extra information!
That's exactly what is going to be checked in `checkEpubHeader()` in https://github.com/IDPF/epubcheck/blob/5c7c4ad01014d71ff7f9ccab3d3519601c1420c1/src/main/java/com/adobe/epubcheck/api/EpubCheck.java#L275

Issue #530, opened 2015-04-29
Error in mimetype file - but it contains right string


robogeek:
I've written a tool, AkashaEPUB, to generate EPUB3's and am using epubcheck as one of the validation tools that I've done the right thing.

The only message being printed is:

```
$ java -jar /usr/bin/epubcheck guide.epub 
Epubcheck Version 3.0.1

ERROR: guide.epub: Mimetype contains wrong type (application/epub+zip expected).
Validating against EPUB version 3.0

Check finished with warnings or errors
```

Funny thing is the mimetype file contains the expected text.

```
$ cat out/mimetype 
application/epub+zip
```

The file has ONLY that text - no line feeds etc.  If I add a line feed, there's another message printed about how the file must ONLY have this mimetype text and nothing else.

The file is written with Node.js code like this:

```
fs.writeFile(path.join(config.root_out, "mimetype"), "application/epub+zip", "utf8", next);
```

In other words, it's telling the OS to make it a UTF-8 file, in case that makes any difference.

Finally, for more detail:

```
$ od -c out/mimetype 
0000000   a   p   p   l   i   c   a   t   i   o   n   /   e   p   u   b
0000020   +   z   i   p
0000024
```

Scratching my head and not seeing the problem.

I've been ignoring the issue - but the Apple iTunes book store is rejecting my EPUB because of this Error.

mattgarrish:
Sounds like you’re probably getting extra headers when you zip the epub. Does the mimetype file name start at byte 30 and the value at byte 38?

 

Matt

robogeek:
Ah...  Here's a dump of the start of the the EPUB contents.  The filename is at byte 30, and that 'K' is at byte 38.  I suppose it's compressed hence the unrecognizable characters.

```
0000000   P   K 003 004 024  \0  \b  \0  \b  \0 366  \b 235   F  \0  \0
0000020  \0  \0  \0  \0  \0  \0  \0  \0  \0  \0  \b  \0  \0  \0   m   i
0000040   m   e   t   y   p   e   K   ,   ( 310 311   L   N   , 311 314
0000060 317 323   O   -   (   M 322 256 312   ,  \0  \0   P   K  \a  \b
0000100   o   a 253   , 026  \0  \0  \0 024  \0  \0  \0   P   K 003 004
0000120 024  \0  \b  \0  \b  \0 366  \b 235   F  \0  \0  \0  \0  \0  \0
[...]

PR #537, opened 2015-05-14
Explain another reason mimetype check can fail.

andrewdotn:
The `checkEpubHeader()` method checks the raw EPUB file for the string
`application/epub+zip` starting at byte 38. If the `mimetype` file happens
to be compressed in the zip file, then the check will fail, but the error
message will only say that its contents are invalid.


tofi86:
related to #530

tofi86:
probably related to #497

andrewdotn:
Helio, any idea when this will be merged in?

tofi86:
Your patch doesn't build so it's likely this will take some time until someone assigns himself to this issue.

andrewdotn:
After updating the warning message in the unit test as well, Travis CI is showing that the code builds and all unit tests pass. Is there something I’m missing?

rdeltour:
LGTM. I'll squash and merge to the master ASAP.

andrewdotn:
Thanks!

PR #595, opened 2015-09-10
fix option name in error message (-mode -> -profile)

takahashim:
I guess this message wrong.

cf. https://github.com/IDPF/epubcheck/blob/master/src/main/java/com/adobe/epubcheck/tool/EpubChecker.java#L589


takahashim:
sorry! already fiexed

tofi86:
Sorry @takahashim, I fixed it straightforward on `master` 😉

takahashim:
@tofi86 thanks for quick fix!

PR #608, opened 2015-09-25
BitmapChecker removes each temporary file after use. Prevent hard drive filling up when launched as an internal library.

francisco-sanchez-molina:
BitmapChecker removes each temporary file after use. Prevent hard drive filling up when launched as an internal library.


yebenes:
:+1: I consider it an interesting feature. Why is not added?

rdeltour:
Rebased and merged as 8dccaca. Thanks @francisco-sanchez-molina!

francisco-sanchez-molina:
thank you so much!

PR #616, opened 2015-10-02
use default variable format in messages

tofi86:
@rdeltour I know that `%1s` is also a valid format for string formatting variables in messages, but it is normally used to create fixed width strings like in `%3s` which would always create a string with 3 characters minimum.

I'd suggest to always use the default variable format `%1$s` instead. Okay for you?


rdeltour:
LGTM, it was probably a typo  :blush:

tofi86:
> LGTM, it was probably a typo :blush:

yeah, that's what I assumed as well... came up as a question over at Transifex from Dutch Translator Merijn...

Waiting for the CI build, then going to merge this...

Issue #584, opened 2015-09-04
HTM-015 reports wrong path and duplicates with HTM-004


tofi86:
The message `HTM-015` reports a wrong path and duplicates with `HTM-004`:

```
ToFi-MBP in ~/Downloads/epubcheck-4.0.0
○ → java -Duser.language=en -Duser.country=US -jar epubcheck.jar /Users/tobiasfischer/Desktop/epub3-xhtml-doctype.epub
Validating using EPUB version 3.0.1 rules.
ERROR(HTM-004): /Users/tobiasfischer/Desktop/epub3-xhtml-doctype.epub/EPUB/lorem.xhtml(-1,-1): Irregular DOCTYPE: found '-//W3C//DTD XHTML 1.1//EN', expected '<!DOCTYPE html>'.
WARNING(HTM-015): /Users/tobiasfischer/Desktop/epub3-xhtml-doctype.epub/lorem.xhtml(-1,-1): HTML4 DOCTYPE definition within EPUB v3.

Check finished with errors

epubcheck completed
```

I would expect EITHER `HTM-004` OR `HTM-015`—but in any case I would expect `HTM-015` to report the correct path to the the HTML file: `epub3-xhtml-doctype.epub/EPUB/lorem.xhtml` instead of `epub3-xhtml-doctype.epub/lorem.xhtml`. This may also apply to `HTM-014`, `HTM-016`, etc. as the `EpubHTML5StructureCheck` class only uses the manifest item `href` attribute:

```java
if ((0 != (docTypeMatches & hasHTML4)) && (epubPackage.getVersion() == EPUBVersion.VERSION_3))
{
  report.message(MessageId.HTM_015, EPUBLocation.create(mi.getHref()));
}
else if ((0 != (docTypeMatches & hasHTML5)) && ((hasXhtml != (docTypeMatches & hasXhtml)))  &&  (epubPackage.getVersion() == EPUBVersion.VERSION_2))
{
  report.message(MessageId.HTM_016, EPUBLocation.create(mi.getHref()));
}
```

Dummy EPUB: https://www.dropbox.com/s/vza90w6f7oelida/epub3-xhtml-doctype.epub?dl=1

tofi86:
Should be `EPUBLocation.create(epubPackage.getPackageMainPath() + "/" + mi.getHref())` instead of `EPUBLocation.create(mi.getHref())` but this breaks a lot of tests so I won't commit...

rdeltour:
[...]

PR #618, opened 2015-10-03
Improve checks of XHTML doctypes

tofi86:
partial fix for #584 
@rdeltour please have a look and merge if it's okay...

The duplicate messages reported in #584 still need to be fixed, that's why I'll leave the ticket open and assigned to you...


tofi86:
Thanks! :)

PR #673, opened 2016-02-13
Generate XML with JVM libraries in javax.xml

### LINKED ISSUE #486: XML output and error handling
The XML output from epubcheck is very constrained. The message elements contain only a plaintext message with the encountered error.

```
<message>
  OPF-003, WARN, [Item 'iTunesMetadata.plist' exists in the ePub, 
  but is not declared in the OPF manifest.], 978XXXXXXXXXX.epub
</message>
```

Currently, automatic error handling is only possible by string analysis. It would be helpful to have attributes besides the error message which provide 
 * error code
 * file (e.g. 'iTunesMetadata.plist', '978XXXXXXXXXX.epub')
 * error type (e.g. 'error', 'warning')
 * source path (optional)

The XML output could be read as follows:

```
<message code="OPF-003" type="warning" sourcepath="/package/manifest[1]">
 Item 'iTunesMetadata.plist' exists in the ePub, 
  but is not declared in the OPF manifest.
</message>
```

### LINKED ISSUE #517: Option for including list of all resources + media types in output [enhancement]
Would it be possible to add an option to include a list of all manifest resources (including their media types) in Epubcheck's output? This would enable one to screen files for the presence of user-defined media types. Currently I'm looking into ways to screen Epubs for foreign resources and DTBook content (neither are well supported by readers). This information is already present in the manifest:

    <manifest>
        <item href="toc.ncx" id="ncx" media-type="application/x-dtbncx+xml" />
        <item href="Images/pdfVenn.jp2" id="pdfVennJP2" media-type="image/jp2" fallback="pdfVennPNG" />
        <item href="Images/pdfVenn.png" id="pdfVennPNG" media-type="image/png" />
        <item href="hauy-2005-1.xml" id="opf3" media-type="application/x-dtbook+xml" />
    </manifest>
   
For some more context to what I'm trying to do here, also have a look at this blog post I wrote about assessing EPUBs against a technical profile using Epubcheck:

[Policy-based assessment of EPUB with Epubcheck](https://researchkb.wordpress.com/2015/03/13/policy-based-assessment-of-epub-with-epubcheck/)

### LINKED ISSUE #670: Illegal characters in XML output
While processing an encrypted EPUB with Epubcheck 4.0.1 (with output to XML format), I ended up with the following output file:

<https://github.com/KBNLresearch/epubPolicyTests/blob/master/epubcheckout/4.0.1/epub20_encryption_binary_content.xml>

Line 17 of the output file contains a warning about an illegal XHTML Named entity. However, this very entity (Unicode code point U+000B / 0xb) is included in the output file (line 17, column 81), which makes the output itself invalid XML!  This creates problems if the XML needs to be processed further down the line (in my case I want to run some Schematron rules on it).

For rights reasons I cannot share the original EPUB,  but I created a synthetic file that reproduces the problem at:

<https://github.com/KBNLresearch/epubPolicyTests/blob/master/build/epub20_encryption_binary_content.epub?raw=true>

(See also about control characters in XML: <https://www.w3.org/International/questions/qa-controls.en.php>)

tledoux:
This commits changes the generation of XML reports to use regular Java
libraries avoiding bad output.
It also checks for not UTF-8 characters and escapes them.
Finally, it adds the list of media-types included in the epub.
As per suggestion of @tofi86, we add the subMessage and severity
attributes to message to get the exact code.

The tests have been enhanced to better compare the actual and
expected results. In particular, the error codes are now checked instead of the messages themself
to be more robust on locales and small syntactic corrections.

Some tests cases have been added to test for encrypted or obfuscated
epubs.

Fixes #670.
Fixes #517.
Fixes #486.


tofi86:
Thanks for working on this, Thomas!

Issue #711, opened 2016-11-06
false positives due to locale settings


dilbirligi:
It seems epubcheck is not handling page-lists as 
https://idpf.github.io/a11y-guidelines/content/nav/pagelist.html
says it should.
I get the following even though i don't think i should:
ERROR(OPF-027): OEBPS/nav.xhtml(91,41): Undefined property: 'page-list'.
ERROR(NAV-003): OEBPS/content.opf(-1,-1): The Navigation Document must have a page list when content document(s) contain page breaks (epub:type="pagebreak").

And all of the examples with page-lists that i've found (such as childrens-literature.epub) also fail with these kinds of errors. I even tried changing type to EDUPUB, but that doesn't improve things (& i think it shouldn't be required to have a page-list in EPUB3 anyway).

Strangely, idpf.org validation says no errors on the same epubs, & it says it also uses epubcheck 4.0.1.

dilbirligi:
The following is my starter page-list (**nav.xhtml**):
```
			<nav epub:type="page-list" hidden="hidden"><!--  id="pages"-->
					 <h2>Pagebreaks of the print version, first edition</h2> 
					 <ol>
								<li>
										  <a href="title-page-i.xhtml#p01">1</a>
								</li>
								<li>
										  <a href="title-page-ii.xhtml#p02">2</a>
								</li>
					 </ol>
			</nav>
```
And the following is source (**content.opf**):
```
			<dc:source id="src">http://www.ebay.com/itm/ULKULERiM-/122208566685?hash=item1c7432bd9d:g:tOIAAOSwHMJYGO6N</dc:source>
		        <meta property="source-of" refines="#src">pagination</meta>
```
And the following is a sample page break (**title-page-i.xhtml**#p01):
`<div class="center"><span class="pb" epub:type="pagebreak" title="1" id="p01">1</span></div>`

P.S. So now i'm faced w/ a decision:
1. use this cause it works (e.g., in ADE; & hopefully epucheck wil eventually be fixed not to give 2 errors about page-list setup for no reason);
OR
2. don't use a page-list at all simply because buggy epubcheck doesn't like it, & i'm an obsessive dude who would like to have no errors or warnings from epubcheck.
P.P.S. What's your take on this, & how i should proceed?
P.P.P.S. I even adjusted (& thus resolved) everything additional related to EDUPUB, but the 2 errors mentioned epubcheck gives for a page-list regardless of type being EPUB or EDUPUB.

[...]

PR #712, opened 2016-11-07
Improves Epub30CheckExpandedTest.testPageList() by actually setting an appropriate default locale for the duration of the test 

dilbirligi:
that tests that the case transformations are indeed locale-independent.

Better testing of fix for issue #711.


tofi86:
Looks like a reasonable test for #711 and 70e990d. Will merge this for next release now that  	70e990d (#711) is already shipped with 4.0.2

Issue #663, opened 2015-12-15
Empty guide in EPUB 2 not detected


pettarin:
Trying to validate an EPUB 2 with an empty `<guide></guide>` in the OPF does not return any error or warning.

Trying to validate an EPUB 3 with an empty `<guide></guide>` in the OPF does return:

```
ERROR(RSC-005): test.epub/OEBPS/content.opf(41,9): Error while parsing file 'element "guide" incomplete; missing required element "reference"'. 
```

The EPUB 2 spec (assuming this: http://www.idpf.org/epub/20/spec/OPF_2.0.1_draft.htm ) clearly implies that the `<guide>` is optional, but, if present, it should have at least one child.

I think the .rng currently in use has a "zeroOrMore" instead of a "oneOrMore", which was brought in by the patch in issue #24 .

So the question is: is an empty `<guide> </guide>` legal in EPUB 2 or is it an epubcheck bug?

rdeltour:
Looks like a bug to me, [OPF 2.0.1 says](http://www.idpf.org/epub/20/spec/OPF_2.0_latest.htm#Section2.6)

> Within the package there may be one guide element, containing one or more reference elements.

Good catch Alberto!

tofi86:
@rdeltour can this be fixed in the EpubCheck schema directly or does it need to be fixed in the official spec schema and only then can be used in EpubCheck?

jajessup:
Tobias, that object is declared in the same file as the public method that returns it. I’m not a java expert but beside IKVM, it didn’t look right having a “Public” method returning a “Private” object.

 

If a “Public” method is returning a object, that object should be declared “Public” also. This just looks like a simple oversight.

santoch:
Yes, it was an oversight.  My apologies.
________________________________

jajessup:
No problem. Thanks for your efforts!

tofi86:
This seems to be fixed in latest revision of the OPF schema at http://www.idpf.org/epub/20/spec/OPF_2.0_latest.htm#AppendixA but hasn't been updated in the schema copy in EpubCheck.

tofi86:
Has been fixed in PR #724.

Issue #114, opened 2013-10-15
Empty dc:identifier values not rejected


rdeltour:
_From [abdela...@gmail.com](https://code.google.com/u/116745328488937130160/) on July 27, 2011 21:33:15_

An OPF containing  <dc:identifier id="bookid" />
does not raise any validation errors in epubcheck 1.2. This blank dc:identifier value
is invalid and should cause validation to fail.

The OPF 2.0.1 spec defines it as """A string or number used to uniquely identify the resource"""

_Original issue: http://code.google.com/p/epubcheck/issues/detail?id=114_

PR #725, opened 2016-12-21
Update OPF 2.0 RNG schema partially to latest version (changing datatype text to anyURI for href attributes)

tofi86:
When working on #663 and #724 I came across the fact, that the OPF RNG schema in EpubCheck is not up to date.

Schema from http://www.idpf.org/epub/20/spec/OPF_2.0_latest.htm#AppendixA differs in several places from the EpubCheck schema in `src/main/resources/com/adobe/epubcheck/schema/20/rng/opf20.rng`.

I'm not familiar with the process of updating official schemas in the EpubCheck project, so how are the analogues DTDs handled? Where do they come from?

Also: it seems for EpubCheck the RNG `<start>` element must not be present in order to be able to run the checker... right?

Note: the initial commit on this PR breaks the build. The OPF 2.0 schema update obviously has side-effects on the tests which needs to be handled... How?
IMHO, this may also be a reason to make the next release version 4.1.0 and not 4.0.3...

### DIFF
--- a/src/main/resources/com/adobe/epubcheck/schema/20/rng/opf20.rng
+++ b/src/main/resources/com/adobe/epubcheck/schema/20/rng/opf20.rng
@@ -18,11 +18,6 @@ Authors:
          Peter Sorotokin <psorotok@adobe.com>
 -->
 
-<!-- must not be present for EpubCheck validation. -->
-<!--<start>
-  <ref name="OPF20.package-element"/>
-</start>-->
-
 <define name="OPF20.optional-id-attribute">
   <optional>
     <attribute name="id">
@@ -71,6 +66,13 @@ Authors:
   </optional>
 </define>
 
+<define name="OPF20.optional-xsi-type">
+  <optional>
+    <attribute name="type" ns="http://www.w3.org/2001/XMLSchema-instance">
+      <text/>
+    </attribute>
+  </optional>
+</define>
 
 <define name="OPF20.package-element">
   <element name="package">
@@ -166,9 +168,12 @@ Authors:
       <data type="ID"/>
     </attribute>
     </optional>
-
+    <ref name="OPF20.optional-xsi-type"/>
     <ref name="OPF20.optional-scheme-attribute"/>
-    <ref name="DC.metadata-common-content"/>
+    <!-- <ref name="DC.metadata-common-content"/>
+    MG 20111024, see epubcheck Issue 114
+    -->
+    <ref name="DC.metadata-required-content"/>    
   </element>
 </define>
 
@@ -183,7 +188,7 @@ Authors:
 <define name="DC.language-element" ns="http://purl.org/dc/elements/1.1/">
   <element name="language">
     <ref name="OPF20.optional-id-attribute"/>
-
+    <ref name="OPF20.optional-xsi-type"/>
     <ref name="DC.metadata-common-content"/>
   </element>
 </define>
@@ -211,7 +216,7 @@ Authors:
     </element>
     <element name="date">
       <ref name="OPF20.optional-id-attribute"/>
-
+      <ref name="OPF20.optional-xsi-type"/>
       <ref name="OPF20.optional-event-attribute"/>
       <ref name="DC.metadata-common-content"/>
     </element>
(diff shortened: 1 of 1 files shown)


src/main/resources/com/adobe/epubcheck/schema/20/rng/opf20.rng
```diff
@@ -168,12 +166,9 @@ Authors:
...
     </attribute>
     </optional>
-    <ref name="OPF20.optional-xsi-type"/>
+
     <ref name="OPF20.optional-scheme-attribute"/>
-    <!-- <ref name="DC.metadata-common-content"/>
-    MG 20111024, see epubcheck Issue 114
-    -->
-    <ref name="DC.metadata-required-content"/>    
+    <ref name="DC.metadata-common-content"/>
```
tofi86: oh, and this may also interfer with #114 

cannot say if the changes from #114 are reflected in the latest schema version from the specs...

tofi86:
@mattgarrish Matt, may I request your review here? I had trouble figuring things out when trying to update the OPF RNG schema to the latest version. See comment above.

mattgarrish:
I'll have a look, but might not be today. One question I can answer quick is that, yes, the start element is not included in the opf20.rng file. That file is imported into opf.rng, which defines the start element.

mattgarrish:
That DTD directory looks like a relic of OEB 1.2, EPUB's precursor. Are they actually used in epubcheck? @rdeltour ?

src/main/resources/com/adobe/epubcheck/schema/20/rng/opf20.rng
murata0204: It appears that the schema from http://www.idpf.org/epub/20/spec/OPF_2.0_latest.htm#AppendixA allows empty dc:identifier elements.  This is because 


```
<define name="DC.metadata-common-content">
  <text/>
</define>
```
allows any text including the empty string.

murata0204:
I compared the schema in the EPUB 2 spec and that in epubcheck.  Yes, there are some differences.  I have a feeling that all of them are required for epubcheck, if we allow empty guide elements.  So, we do not have to use the latest schema from the spec.

mattgarrish:
>  I have a feeling that all of them are required for epubcheck

I think it might be a mix. Some of the changes look like reversions. Why has xsi:type been removed, for example? OPF 2.0.1 says:

> For compatibility with Guidelines for implementing Dublin Core in XML (http://dublincore.org/documents/dc-xml-guidelines/) this specification allows xsi:type attribute for metadata items that can be given using some sort of encoding scheme and xml:lang attribute when an item can be given using human-readable text. Elements that allow xsi:type attribute are identifier, language, date, format and type.

(But, admitting again that I've only had a chance to skim.)

murata0204:
I guess that some people use XSD validation anyway, and thus the RNG schema in epubcheck was forced to allow xsi:type.  But even if EPUB  2.0.1 intentionally dropped this attribute, I do not think that we should change epubcheck at this stage.  Somebody might be using it heavily.

mattgarrish:
> even if EPUB 2.0.1 intentionally dropped this attribute

And that's the thing, it didn't.

From what I can tell, epubcheck's schema doesn't match the one in 2.0.1 because it's been updated over time to fix problems in the hard-coded appendix version. I believe that's also why we chose to link to the NVDL files from the spec in 3.0, so that we can update the RNC/Schematron rules without needing revisions. (Granted, I don't know that we're being good at submitting patches back to the "official" schemas.)

So, on review, I'd suggest reverting almost all the changes not requested by an issue. In other words:

- erase the start element comment
- add back the xsi:type attribute
- keep using DC.metadata-required-content for dc:identifier
- keep the restriction against openebook elements (not sure why this was added, but must have been an issue for someone)

Changing text to anyURI for href attributes looks like a good fix to me, though. So keep that and the requirement for guide not to be empty.

tofi86:
Okay, so I rebased on current `master` which resolves the `<guide>` issue as this has alreday been handled with #724 to fixed #663 

So we'll only keep changing `text` to `anyURI` for `@href` attributes

tofi86:
@mattgarrish if you approve these changes I'm gong to merge it.

mattgarrish:
Sounds good to me.

I looked at the 2.0 schema and it had the xsi:type definition. If someone intended to drop the attribute from 2.0.1 they forgot to do it in the spec, which means it's still normatively allowed.

tofi86:
Merged, Thanks!

PR #650, opened 2015-11-12
Flexible localization options.

matthew-macgregor:
## Summary

The existing localization code relies upon Locale.getDefault(), which uses the Locale of the host JVM. While this can be changed programmatically, the change applies globally for the JVM. This is fine for client-side use, but it presents issues when used in a server-side environment where each request may need to use a different locale concurrently. 

After attempting several workarounds and investigating code, I realized that this use case hasn't been addressed and decided to attempt a solution.

You can see a discussion here:

[StackOverflow: How to change the Locale of epubcheck when using it as a library](http://stackoverflow.com/questions/33586984/how-to-change-the-locale-of-epubcheck-when-using-it-as-a-library/33588298)
### Goals
- Allow clients to pass a locale to the EpubCheck class.
- Fall back to the default locale if no explicit locale is provided.
- Preserve caching of the resource-hungry localized strings.
- Use lazy initialization of messages wherever possible. 
- Minimize disturbance of top-level interfaces.
### Usage

Usage of the new functionality is simple:

```
Locale l = new Locale("es");
EpubCheck epubCheck = new EpubCheck(file);
epubCheck.setLocale(l); // new
epubCheck.doValidate();

// Validación usando la versión de reglas epub 2.0.1.
```
### Changes
- Refactored `com.adobe.epubcheck.util.Messages` to cache an instance of itself for each locale, as they are requested. Updated all the tests.
- Refactored `com.adobe.epubcheck.messages` to use a similar pattern. The overridden messages added some complications. As an attempt to modularize, I broke `MessageDictionary.java` into several classes, which allowed me to cache what I wanted to cache (default messages and severities) while keeping the override functionality intact. 
- I added `setLocale()` to the `Report` interface. This will break any clients that are passing custom reports. The code works without this with minor modifications (since we have access via the `MasterReport`) but it seemed more consistent to keep this on the interface. It could easily be removed to avoid the breaking change. 
- Updated tests and added new tests for the additional functionality. 
### Notes & Limitations
- There are currently two Message API's in the code base: one which handles epubcheck-related messages and the other which handles "utility" messages. I left this division intact, but propose that the util messages should be merged into the more robust MessageDictionary structure in some way.
- Since localization is tied to the MasterReport, clients that provide a custom Report object will find it difficult to benefit from localization without inheriting from MasterReport. This is a problem with both the existing code and my changes. I didn't address it.

Thanks very much, I'm looking forward to feedback.

PR #731, opened 2016-12-28
Add better url checks

### LINKED ISSUE #708: Enhanced URL checks
When having an invalid link / uri in your _(EPUB 2)_ XHTML data:

```
<a href="https:/www.youtube.com/watch?v=xxxxxxxxxxx">Youtube link</a>
```

EpubCheck 3.0.1 throws the following `WARNING`:

```
WARNING: 123456789.epub/OEBPS/kapitel1.xhtml(15,62): use of non-registered URI scheme type in href: https:/www.youtube.com/watch?v=xxxxxxxxxxx
```

But EpubCheck 4.0.1 does not!

The error message `OPF_021`

```
OPF_021=Use of non-registered URI scheme type in href: '%1$s'.
```

is still present, but the warning only seems to get thrown when validating DTBook data (https://github.com/IDPF/epubcheck/blob/master/src/main/java/com/adobe/epubcheck/dtbook/DTBookHandler.java#L99)

It seems the check was accidentally removed during the 4.0 refactoring?!

I expect this also to happen with EPUB 3, but haven't tested yet...

tofi86:
Add new Warning(RSC-023) when host of HTTP-URL is null (fixes #708).

Catches URL typos like `https:/www` or `https:www`

----

Plus: also report HTTPS URLs in the list of references (c3c9204)

### DIFF
--- a/src/main/resources/com/adobe/epubcheck/messages/MessageBundle.properties
+++ b/src/main/resources/com/adobe/epubcheck/messages/MessageBundle.properties
@@ -293,6 +293,7 @@ RSC_019=EPUBs with Multiple Renditions should contain a META-INF/metadata.xml fi
 RSC_020='%1$s' is not a valid URI.
 RSC_021=A Search Key Map Document must point to Content Documents ('%1$s' was not found in the spine).
 RSC_022=Cannot check image details (requires Java version 7 or higher).
+RSC_023='%1$s' is not a valid URL.
 
 #Scripting
 SCP_001=Use of Javascript eval() function in EPUB scripts is a security risk.
--- a/src/main/resources/com/adobe/epubcheck/messages/MessageBundle.properties
+++ b/src/main/resources/com/adobe/epubcheck/messages/MessageBundle.properties
@@ -293,7 +293,7 @@ RSC_019=EPUBs with Multiple Renditions should contain a META-INF/metadata.xml fi
 RSC_020='%1$s' is not a valid URI.
 RSC_021=A Search Key Map Document must point to Content Documents ('%1$s' was not found in the spine).
 RSC_022=Cannot check image details (requires Java version 7 or higher).
-RSC_023='%1$s' is not a valid URL.
+RSC_023=The URL '%1$s' is missing %2$d slash(es) '/' after the protocol '%3$s:'
 
 #Scripting
 SCP_001=Use of Javascript eval() function in EPUB scripts is a security risk.
--- a/src/main/java/com/adobe/epubcheck/messages/MessageDictionary.java
+++ b/src/main/java/com/adobe/epubcheck/messages/MessageDictionary.java
@@ -329,6 +329,7 @@ public class MessageDictionary
       map.put(MessageId.RSC_020, Severity.ERROR);
       map.put(MessageId.RSC_021, Severity.ERROR);
       map.put(MessageId.RSC_022, Severity.INFO);
+      map.put(MessageId.RSC_023, Severity.WARNING);
 
       // Scripting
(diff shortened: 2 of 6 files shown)


tofi86:
@rdeltour can you do a quick review soon as this blocks #650 rebase?

src/main/resources/com/adobe/epubcheck/messages/MessageBundle.properties
```diff
@@ -293,6 +293,7 @@ RSC_019=EPUBs with Multiple Renditions should contain a META-INF/metadata.xml fi
 RSC_020='%1$s' is not a valid URI.
 RSC_021=A Search Key Map Document must point to Content Documents ('%1$s' was not found in the spine).
 RSC_022=Cannot check image details (requires Java version 7 or higher).
+RSC_023='%1$s' is not a valid URL.
```
rdeltour: I wouldn't say "is not a valid URL", since technically it is (it's just that the host component is empty).
What about something more explicit along the lines of: `URL '%1$s' has no host component`? (or, maybe more user friendly.

tofi86: Yeah, you're right, that's technically wrong. I was trying to phrase this more user friendly.

`URL '%1$s' has no host component` is technically correct, but probably hard to understand for non-geeks... ;-)

If we phrase it more technical I would like to add something like `…component. Probably misspelled or missing http(s):// ?`. But I'm not sure whether this makes for a good warning message...

rdeltour: I agree, not sure what's the most understandable option... (my useless 2 cents ;-)

tofi86: Okay, technically it's not the host component which is missing. The `URI` parser just can't identify it.

Technically it's just a missing `/` or two. That's the only cases where `RSC_023` is triggered. For `https:/www.yout` and for `https:www.yout`.

All the other cases are covered by RSC_020 or HTM_025.

So I suggest we just say exactly that: There's a missing slash after the protocol/schema of the URL. protocol is probably more known to non-geeks, right?

```
RSC_023=The URL '%1$s' is missing one or two slashes '/' after the protocol '%2$s:'
> The URL 'https:/www.youtube.com' is missing one or two slashes '/' after the protocol 'https:'
```

We could even go further and say:

```
RSC_023=The URL '%1$s' is missing %2$d slash(es) '/' after the protocol '%3$s:'
> The URL 'https:/www.youtube.com' is missing 1 slash(es) '/' after the protocol 'https:'
```

What do you think?

tofi86:
> Okay, technically it's not the host component which is missing. The URI parser just can't identify it.
> 
> Technically it's just a missing / or two. That's the only cases where RSC_023 is triggered. For https:/www.yout and for https:www.yout.
> 
> All the other cases are covered by RSC_020 or HTM_025.
> 
> So I suggest we just say exactly that: There's a missing slash after the protocol/schema of the URL. protocol is probably more known to non-geeks, right?
> 
> RSC_023=The URL '%1$s' is missing one or two slashes '/' after the protocol '%2$s:'
> > The URL 'https:/www.youtube.com' is missing one or two slashes '/' after the protocol 'https:'
> 
> We could even go further and say:
> 
> RSC_023=The URL '%1$s' is missing %2$d slash(es) '/' after the protocol '%3$s:'
> > The URL 'https:/www.youtube.com' is missing 1 slash(es) '/' after the protocol 'https:'
> 
> What do you think?

I rephrased the message according to my last suggestion.

@rdeltour what do you think? are you okay with that?

PR #732, opened 2016-12-28
[fix #665] Set -Djava.awt.headless=true in EpubChecker class

### LINKED ISSUE #665: GUI focus stolen due to epubcheck launching window-less “Checker” empty GUI app
Running epubcheck from the command-line still launches a window-less GUI app named Checker that claims the GUI’s focus. The app stays open until epubcheck finishes its thing, at which point it kills itself (returning the focus it stole if the user did not claim it back).

It thus becomes impossible to run epubcheck in the background while the user works on something else, specially when running batch-checks of lots of EPUBs (which involves calling epubcheck once per ebook, therefore causing that dreaded GUI app to launch repeatedly), since the user is then constantly taken away from whatever app he or she is using.

Observed on OS X Yosemite 10.10.5 (14F27) and JRE 1.8.

tofi86:
Set `-Djava.awt.headless=true` in `EpubChecker` class to fix #665 (_GUI focus stolen due to epubcheck launching window-less “Checker” empty GUI app_)

Needs to be set as early as possible in a static way.

### DIFF
--- a/src/main/java/com/adobe/epubcheck/tool/EpubChecker.java
+++ b/src/main/java/com/adobe/epubcheck/tool/EpubChecker.java
@@ -62,6 +62,13 @@ import com.adobe.epubcheck.util.outWriter;
 public class EpubChecker
 {
 
+  static {
+    /* fix #665 (window-less "Checker" gui app on Mac)
+     * set -Djava.awt.headless=true programmatically as early as possible
+     */
+    System.setProperty("java.awt.headless", "true");
+  }
+
   String path = null;
   String mode = null;
   EPUBProfile profile = null;


src/main/java/com/adobe/epubcheck/tool/EpubChecker.java
```diff
@@ -62,6 +62,13 @@
 public class EpubChecker
 {
 
+  static {
+    /* fix #665 (window-less "Checker" gui app on Mac)
+     * set -Djava.awt.headless=true programmatically as early as possible
+     */
+    System.setProperty("java.awt.headless", "true");
```
rdeltour: Could also have been added to the static main method in `Checker`, but as far as I can tell this is OK too.

PR #737, opened 2017-01-02
[fix #540] Include file path info in uri field in XML output

### LINKED ISSUE #540: Include file path info in uri field
Earlier this week I was running epubcheck on a large nested directory tree. While looking at some of the output I was surprised that the _uri_ field only gives the file name without its path. E.g.: 

```
<repInfo uri="whatever.epub">
```

Wouldn't it be better to include the full file path here as well (particularly helpful when working with directory trees)?

This applies to both 3.0.1 and the 4.0 pre-release.

tofi86:
fixes #540

However, as this may break existing implementations which retrieve the filename from the `uri` attribute, it's probably better suited for the next minor release instead of the next bugfix release.

What do you think, @rdeltour?


tofi86:
@tledoux also wanted to take a look at the previous merge conflicts. But now that you took action and tried merging current master to this branch which still has test errors, will you take over and fix this?

tledoux:
@tofi86 indeed there seems to be conflicting with the last merge. I try to push the fix.

tledoux:
Last fix looks good to me (conflict with last mods and working on Windows).

Feel free to merge then.
Thanks @tofi86

tofi86:
Done. Thanks for having a look and for fixing conflicts @tledoux 👍

PR #741, opened 2017-01-04
[fix #740] Query report message counter to fix wrong exit message for single file validation

### LINKED ISSUE #740: Single file validation always exits with ERROR also when there are only WARNINGs
When validating a single file like

```
--mode nav -v 3.0 /Users/tofi/Entwicklung/eclipse-workspace/epubcheck/src/test/resources/30/single/nav/invalid/nav-landmarks-duplicates.xhtml
```

the result could be as follows:

```
WARNING(RSC-017): ./src/test/resources/30/single/nav/invalid/nav-landmarks-duplicates.xhtml(28,64): Warning while parsing file 'Duplicate 'a' elements with the same 'epub:type' and 'href' attributes inside 'landmarks' nav element'.
WARNING(RSC-017): ./src/test/resources/30/single/nav/invalid/nav-landmarks-duplicates.xhtml(30,64): Warning while parsing file 'Duplicate 'a' elements with the same 'epub:type' and 'href' attributes inside 'landmarks' nav element'.

Check finished with errors

epubcheck completed
```

We get the message `Check finished with errors` but do not have any ERROR's here, only WARNINGs.

This is because the single file validation is based on `DocumentValidator` interface whose `validate()` method does not return a status code but a boolean value whch does not allow for separation between Errors and Warnings.

Alternatively we could query the current `Report` and get Error/Warning/etc. count from it with `report.getXXXCount()` and decide upon this what message should be issued.

Implementing this could also avoid code duplication in the current if/else clause.

What do you think about this, @santoch? I think it was you who introduced this with the B&N contribution?

tofi86:
fixes #740

### DIFF
--- a/src/main/java/com/adobe/epubcheck/tool/EpubChecker.java
+++ b/src/main/java/com/adobe/epubcheck/tool/EpubChecker.java
@@ -174,14 +174,23 @@ public class EpubChecker
     }
     else
     {
-      if (check.validate())
+      boolean validationResult = check.validate();
+      if (validationResult)
       {
         outWriter.println(Messages.get("no_errors__or_warnings"));
         return 0;
       }
-      System.err.println(Messages.get("there_were_errors"));
+      else if (report.getWarningCount() > 0 && report.getFatalErrorCount() == 0 && report.getErrorCount() == 0)
+      {
+        System.err.println(Messages.get("there_were_warnings"));
+        return failOnWarnings ? 1 : 0;
+      }
+      else
+      {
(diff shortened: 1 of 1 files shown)


tofi86:
Querying the `report` shouldn't be a problem, as it's already done in `EpubCheck.doValidate()` to determine the `int` return value...

tofi86:
Does not break any API or such, just a bugfix. I'm confident about it, so I'm going to merge it by my self.

PR #742, opened 2017-01-05
[fix #655] display error/warning count in EpubCheck result

### LINKED ISSUE #655: error/warning count in output
I was just doing some work on generating an epub from the specs and got to thinking it would be nice to see the total error/warning count after the message that the file contains errors.

When I'm tweaking xslts, it's not always clear if I've fixed the problem without scrolling back through all the messages output to the console. A count would at least give me a quick visual whether things have gotten better, worse or stayed the same.

Anyway, another of my non-critical feature requests...

tofi86:
fixes #655

To be able to implement this I had to change the `Report` interface and add two new methods `getInfoCount()` and `getUsageCount()`.

However, this change may break existing Java implementations using the plain `Report` interface instead inheriting from `MasterReport`.

So this definately needs to be in a Minor release rather than a Bugfix release.

### DIFF
--- a/src/main/java/com/adobe/epubcheck/api/MasterReport.java
+++ b/src/main/java/com/adobe/epubcheck/api/MasterReport.java
@@ -19,7 +19,7 @@ import com.adobe.epubcheck.util.ReportingLevel;
 public abstract class MasterReport implements Report
 {
   public static Set<MessageId> allReportedMessageIds = new HashSet<MessageId>();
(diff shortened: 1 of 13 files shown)


tofi86:
I'm confident with those changes and will merge them to master. No big changes anyways, but an Interface change (labeled it accordingly) which is okay since we release v4.1 with other Report interface changes as well...

PR #750, opened 2017-03-04
[fix #680] fix incorrect warning ACC_011

### LINKED ISSUE #680: Incorrect warnings for svg in epubcheck 4.0.1
Now that Sigil has added support for epub3, I have started to receive bugs related to user's epub3 ebooks not passing epubcheck and asking questions about what the "warnings" are for.

According to a Sigil bug report, the following warning -  ACC_011 is reported by epubcheck 4.0.1 (via Pagina) for this snippet of code.

"SVG hyperlinks should have a human-readable title (using the 'title' child element or the 'xlink:title' attribute)"

```
 <div class="svg_outer">
    <div class="svg_inner">
      <svg height="94%" version="1.1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 1200 1800" preserveAspectRatio="xMidYMid meet" width="94%" xmlns:xlink="http://www.w3.org/1999/xlink"><image height="1800" id="title-original" width="1200" xlink:href="../Images/walden-titlepage.jpg"/></svg>
  </div>
</div>
```

After searching the epubcheck code for ACC_011 I see it when handling svg "anchor" tags not svg image tags.  So I am not sure how this warning is being generated.  And according to:
https://developer.mozilla.org/en-US/docs/Web/SVG/Element/image the list of attributes allowed does not even include an  "xlink:title" attribute on that tag to match the xlink:href attribute.  Also according to one of the bug reports, adding xlink:title="blah" to the svg image tag does _not_ prevent epubcheck from generating that warning.  In fact, one of the bug reports states:

```
I have the same error message in one of my epub. 
I have two inline svg, no links inside, no links in the whole xhtml file.
I've tried every possible variations, with "title", "xlink:title", tried to put them inside/outside/between the svg tag, the path tag, etc., no luck... Always the same warning.
It's just a warning but still... It's irritating :-)
```

It almost appears that epubcheck does not recognize the xlink:title tag when it is placed in the svg image or svg main tag.  The svg snippet above is typically how epub ebooks add cover images so it is found in many many places.  epub2 epubcheck does not indicate any problem with it, only epub3. In fact in one book a user found over 200 of these warnings.

So is this a bug in epubcheck 4.0.1?  Should there be a "warning" generated for svg image tags?  If so, why doesn't adding an xlink:title attribute to the svg image tag prevent the warning? How should the author address this warning?

takahashim:
I hope this patch fixes the bug https://github.com/IDPF/epubcheck/issues/680#issuecomment-195761262

### DIFF
--- a/src/main/java/com/adobe/epubcheck/ops/OPSHandler30.java
+++ b/src/main/java/com/adobe/epubcheck/ops/OPSHandler30.java
@@ -679,6 +679,10 @@ public class OPSHandler30 extends OPSHandler
     {
       inRegionBasedNav = false;
     }
+    else if (name.equals("svg"))
+    {
+      inSvg = false;
+    }
   }
 
   /*
--- a/src/test/resources/30/single/xhtml/valid/svg-links.xhtml
+++ b/src/test/resources/30/single/xhtml/valid/svg-links.xhtml
@@ -23,5 +23,8 @@
 				></rect>
 			</a>
 		</svg>
+        <p>
+          <a href="doc.xhtml">another link outside svg</a>
+        </p>
 	</body>
 </html>


rdeltour:
thanks @takahashim 👍

PR #762, opened 2017-06-05
[fix #761] Update ncx.rng schema to add fixed list of pageTarget type values

### LINKED ISSUE #761: ncx pageTarget type not checked
As noted in the samples repository issues (https://github.com/IDPF/epub3-samples/issues/11), the type attribute is not restricted to the values front, normal and special.

The rng definition is currently the very permissive:

```
			<attribute name="type">
				<text/>
			</attribute>
```

But should be an enumeration:

```
			<attribute name="type">
				<choice>
				    <value>front</value>
				    <value>normal</value>
				    <value>special</value>
				</choice>
			</attribute>
```

tofi86:
fixes #761


tofi86:
Thanks for taking a look, @mattgarrish! 👍

PR #763, opened 2017-06-05
[fix #759] fixes nav doc validation to be more faithful to the spec

### LINKED ISSUE #759: navigation doc nav elements without epub:type restricted
According to the specification, only nav elements with an epub:type attribute have to follow the markup restrictions, but I noticed today that nav elements without epub:type generate errors if they don't follow the pattern.

mattgarrish:
- only restrict a nav element's content model when it carries an epub:type attribute
- require a first-child heading for all nav elements not defined in the spec (including without an epub:type)
- require at least one li child for ol (also fixes existing errors in test docs)
- add warning that nav elements should have an epub:type attribute


mattgarrish:
For reference, this PR addresses issue #759

mattgarrish:
Technically it has to be a warning as the other nav elements section says:

> EPUB Navigation Documents optionally MAY include one or more nav elements in addition to the toc, page-list and landmarks nav elements defined above. Such additional nav elements SHOULD have an epub:type attribute to provide a machine-readable semantic, and MUST have a human-readable heading as their first child.

It's trying to encourage people to use semantics in the document because the nav elements are supposed to be machine-processable to some degree.

I don't really like it, either, though, as I think it breaks the case that made me start all this, and sort of forces you to the restricted content model. I think we'd have to open a spec issue to drop it to "may have".

mattgarrish:
Actually, the more I look at the specification, the more flaws I see in the navigation document design. It says that reading systems must present the links, but that's not why the landmarks are there for example. They facilitate behaviours, if such behaviours are supported. I don't know if any reading system gives blanket access to all the nav elements except when the document is in the spine, either.

It also says that a reading system must relocate the reading position, but we clarified in 3.1 that the navigation document can have lists of links to external resources.

Not that the above are directly relevant to interpreting what 3.0.1 says, but the design is showing some cracks.

I was also assuming that a nav without an epub:type was completely detached from processing before digging in, as the nav document is also supposed to be a fully-functional content document allowing any kind of content.

Some more thought as to the purpose of epub:type in the nav document, when it is required and how nav elements are used might be needed, so I'm okay with taking the epub:type test out.

tofi86:
> It's trying to encourage people to use semantics in the document because the nav elements are supposed to be machine-processable to some degree.

Okay, got it, thanks!

> Some more thought as to the purpose of epub:type in the nav document, when it is required and how nav elements are used might be needed, so I'm okay with taking the epub:type test out.

But I think your proposed changes now reflect the current spec much better than before, so I'd vote for leaving the epub:type tests in!

mattgarrish:
I opened three issues last night in the epub revision repository around this (https://github.com/w3c/publ-epub-revision/issues/975, https://github.com/w3c/publ-epub-revision/issues/976, https://github.com/w3c/publ-epub-revision/issues/977), so maybe in the future the requirements will change, but there's no real arguing that it's not what the spec says right now.

PR #765, opened 2017-06-06
[fix #688] Change style/@type to optional in the SVG Schema/DTD (bugfix)

### LINKED ISSUE #688: SVG type attribute not required on style element
A question was raised on the forums today about why epubcheck reports an error of a missing type attribute on a style element.[1]

The specification language is impossible to decipher, but I found a thread on the SVG list asking the same, and the response is that it appears to be an error in the DTD.[2]

It doesn't appear that an errata or new DTD were ever issued, but it does look we should be less strict as there is a similar erratum about font-face-src being incorrectly required in the DTD. The attribute is more clearly marked as optional in SVG 2.

[1] http://idpf.org/forum/topic-3058
[2] https://lists.w3.org/Archives/Public/www-svg/2005Jan/0096.html

tofi86:
fixes #688

### DIFF
--- a/src/main/resources/com/adobe/epubcheck/schema/20/dtd/svg11.dtd
+++ b/src/main/resources/com/adobe/epubcheck/schema/20/dtd/svg11.dtd
@@ -2272,10 +2272,16 @@
 
 <!ENTITY % SVG.style.attlist "INCLUDE" >
 <![%SVG.style.attlist;[
+<!--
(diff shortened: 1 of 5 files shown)


mattgarrish:
LGTM

tofi86:
Thanks. Can you formally approve the changes with the GitHub functionality so I can merge?

PR #784, opened 2017-08-15
[fix #303] Check MimeType uncompressed

### LINKED ISSUE #303: Not checking mimetype is uncompressed in Zip
_From [adam.ret...@googlemail.com](https://code.google.com/u/117632015505805793587/) on September 17, 2013 13:12:06_

PROBLEM: The epubcheck tool does not check that the mimetype entry in the Zip file is uncompressed, i.e. it is not following this part of the spec - http://www.idpf.org/epub/30/spec/epub30-ocf.html#sec-zip-container-mime

RESULT: At the moment epubs where the mimetype is compressed in the Zip erroneously pass the epubcheck tool 

EXPECTED: If the mimetype is compressed in the epub, the epubcheck tool should fail the validation.

_Original issue: http://code.google.com/p/epubcheck/issues/detail?id=304_

takahashim:
* To check that mimetype file in EPUB is not comressed, add a check in `checkEpubHeader()`
* to avoid checking twice, ignore some checks in `checkTrailingSpaces()`

For EPUB files in https://github.com/IDPF/epubcheck/issues/303#issuecomment-260916592 , epubcheck with this patch finds 1 error in epub30-spec_compressed.epub , but no errors in epub30-spec_uncompressed.epub.

### DIFF
--- a/src/test/java/com/adobe/epubcheck/ocf/OCFCheckerTest.java
+++ b/src/test/java/com/adobe/epubcheck/ocf/OCFCheckerTest.java
@@ -394,24 +394,6 @@ public class OCFCheckerTest
   }
 
   @Test
-  public void testInvalidLoremMimetype30()
-  {
-    ValidationReport testReport = testOcfPackage("/30/expanded/invalid/lorem-mimetype/",
-        EPUBVersion.VERSION_3);
-    if (1 != testReport.getErrorCount() || 0 != testReport.getWarningCount())
-    {
-      outWriter.println(testReport);
-    }
-
-    List<MessageId> errors = new ArrayList<MessageId>();
-    Collections.addAll(errors, MessageId.PKG_007);
-    assertEquals(errors, testReport.getErrorIds());
-    assertEquals(0, testReport.getWarningCount());
-
-    assertTrue(testReport.hasInfoMessage("[format version] 3.0.1"));
-  }
-
-  @Test
   public void testInvalidLoremPoster30()
   {
     ValidationReport testReport = testOcfPackage("/30/expanded/invalid/lorem-poster/",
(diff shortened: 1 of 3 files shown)


tofi86:
Thanks @takahashim! This has been on my list for months but I just didn't find the time for it 🙈

src/test/java/com/adobe/epubcheck/ocf/OCFCheckerTest.java
```diff
@@ -394,24 +394,6 @@ public void testInvalidLoremForeign30()
...
-
-    List<MessageId> errors = new ArrayList<MessageId>();
-    Collections.addAll(errors, MessageId.PKG_007);
-    assertEquals(errors, testReport.getErrorIds());
-    assertEquals(0, testReport.getWarningCount());
-
-    assertTrue(testReport.hasInfoMessage("[format version] 3.0.1"));
-  }
-
-  @Test
```
takahashim: you're right, as I commented in commit log c6cae6daaa130ab892c73effcf1ef18fb85cdb56, `testValidateEPUBPLoremMimetype()` in src/test/java/com/adobe/epubcheck/api/Epub30CheckExpandedTest.java  do test against the same expanded EPUB folder (src/test/resources/30/expanded/invalid/lorem-mimetype/) and checck an error `MessageId.PKG_007`.

https://github.com/IDPF/epubcheck/blob/c1808ac4f7e2fcb61e0f30c204322278864964db/src/test/java/com/adobe/epubcheck/api/Epub30CheckExpandedTest.java#L101-L106

tofi86: Okay, perfect! Just wanted to be sure about that!

PR #787, opened 2017-08-16
add a test for PKG_005

takahashim:
zip64 format file uses extra field of zip format, so lorem-zip64.epub should be invalid.


tofi86:
Great, thanks for adding this additional test! 🎉

PR #792, opened 2017-09-05
[WIP] bad-mediatype-for-ncx.epub (formerly: Issue21.epub)

WSchindler:
Relevant are only 7c7abd2 (bad-mediatype-for-ncx.epub), a refurbishment of Issue21.epub based on Dave's minimal epub, and perhaps f2f4bd7 (foo.text) which explains this in text format.

Though I have managed to create the intended directory structure in my local fork, I couldn't create a clean pull request that only contains what it should and puts that into the right position in IDPF/epubcheck repo.


rdeltour:
Closing this PR, which was part of an early attempt at porting the test suite (since then entirely ported to a new testing framework)

PR #794, opened 2017-09-17
Updating Travis config to use openjdk7 instead of oraclejdk7 which isn't supported by Travis anymore

tofi86:
to use deprecated trusty image after [failures](https://travis-ci.org/IDPF/epubcheck/jobs/276550358)

https://blog.travis-ci.com/2017-08-29-trusty-image-updates

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+group: deprecated-2017Q3
 language: java
 jdk:
   - oraclejdk8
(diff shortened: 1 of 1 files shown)


tofi86:
Looks like Travis cannot run `oraclejdk7` anymore:
* https://github.com/travis-ci/travis-ci/issues/7019#issuecomment-316769716

due to
* http://www.webupd8.org/2017/06/why-oracle-java-7-and-6-installers-no.html

tofi86:
Looks like openjdk7 is only supported on the latest trusty images:
* https://github.com/travis-ci/travis-ci/issues/7019#issuecomment-316769716

tofi86:
@rdeltour FYI. Travis does not support the `oraclejdk7` jdk anymore. We have to use the `openjdk7` image from now on.

Issue #669, opened 2016-01-12
epubcheck 4 doesn't compare identifier in OPF and NCX anymore


tofi86:
When checking EPUB 2 eBooks, epubcheck 4[.0.1] doesn't compare the identifier value in OPF and NCX anymore.

This is the result in epubcheck 3.0.1:

```
> java -jar epubcheck-3.0.1.jar "C:\Users\Tobias Fischer\Desktop\epub2.epub"
Epubcheck Version 3.0.1

Validating against EPUB version 2.0
WARNING: C:/Users/Tobias Fischer/Desktop/epub2.epub/EPUB/lorem.ncx: meta@dtb:uid content 'urn:uuid:550e8400-e29b-41d4-a716-4466674412314--edit' should conform to unique-identifier in content.opf: 'urn:uuid:550e8400-e29b-41d4-a716-4466674412314'

Check finished with warnings or errors
```

And this is the result of epubcheck 4[.0.1]:

```
> java -Duser.language=en -jar epubcheck.jar "C:\Users\Tobias Fischer\Desktop\epu
b2.epub"
Validating using EPUB version 2.0.1 rules.
No errors or warnings detected.
epubcheck completed
```

I understand that comparing identifiers in EPUB 3 documents isn't required anymore, but I think the WARNING should still be issued when checking EPUB 2 files...

tofi86:
Demo file: https://www.dropbox.com/s/kemyc9j3zgrpxbj/epubcheck-669_epub2-invalid.epub?dl=0

vincent-gros:
@tofi86, @rdeltour 
The NCX file should not be check the same way in EPUB 2 and EPUB 3?
Why comparing identifiers is not required anymore in EPUB 3 files?

tofi86:
Okay, this is still working! It _IS_ currently reported, but only when you run EpubCheck with ReportLevel USAGE (commandline argument `--usage`):

    USAGE(NCX-001): /Users/tofi/Downloads/epubcheck-669_epub2-invalid.epub/EPUB/lorem.ncx(-1,-1): NCX identifier ('urn:uuid:550e8400-e29b-41d4-a716-4466674412314--2') does not match OPF identifier ('urn:uuid:550e8400-e29b-41d4-a716-4466674412314').

This behaviour was changed by @rdeltour with cc349b0571cd008009f1f3ff4ad56d1e5e1090e5 to fix #329 and he set the severity to USAGE intentionally:

[...]

Issue #138, opened 2013-10-15
Epub with empty dc:title element raise no warnings


rdeltour:
_From [ned...@gmail.com](https://code.google.com/u/107953480010967884658/) on November 14, 2011 17:35:00_

PROBLEM: Attached is an epub with an empty dc:title element in content.opf.  The OPF 2.0.1 spec states the element is required, but doesn't say it can't be empty.  The EPUB 3 spec makes clear that the title must be at least one character long.

**Attachment:** [notitle.epub](http://code.google.com/p/epubcheck/issues/detail?id=138)

_Original issue: http://code.google.com/p/epubcheck/issues/detail?id=138_

PR #797, opened 2017-09-27
[Cleanup] remove duplicate tests for OPF_055

tofi86:
only the original test issue remains
- testValidateEPUB_issue138
- refs #138


tofi86:
I stumbled upon this issue when I tried to fix #669 because some of the tests for #669 also reported `OPF_055` messages which are not related to #669. So I started this cleanup which took me ~30min

PR #801, opened 2017-11-02
Update list of contributors in README.md and add a maintenance alert

tofi86:
As discussed in todays EpubCheck TaskForce meeting.


tofi86:
Also updated the dist readme file.

Let me know if the new credits list and the maintenance alert works for you or if you want to change something.

tofi86:
Added a couple of changes from @mattgarrish's PullRequest #802.

mattgarrish:
LGTM!

PR #807, opened 2017-11-15
[WIP] [fix #702] Validate content of <dc:language>

### LINKED ISSUE #702: Validate content of <dc:language>
epubcheck doesn't check dc:language value!

According with [specification](http://www.idpf.org/epub/30/spec/epub30-publications.html#elemdef-opf-dclanguage)

Every metadata section must include at least one language element with a value conforming to [RFC5646].

> The following example shows a Publication is in U.S. English.
> <metadata xmlns:dc="http://purl.org/dc/elements/1.1/">
>     …
>     <dc:language>en-US</dc:language>
>     …
> </metadata>

content.opf of my ePUB after export from Adobe InDesign

```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<package xmlns="http://www.idpf.org/2007/opf" xmlns:dc="http://purl.org/dc/elements/1.1/" unique-identifier="bookid" version="2.0">
	<metadata>
		<meta name="generator" content="Adobe InDesign"/>
		<meta name="cover" content="xxx-cover.jpg"/>
		<dc:title>xxxx</dc:title>
		<dc:creator>xxx</dc:creator>
		<dc:subject></dc:subject>
		<dc:description>xxx</dc:description>
		<dc:publisher>Editorial Presença</dc:publisher>
		<dc:date>2016-02-11</dc:date>
		<dc:source></dc:source>
		<dc:relation></dc:relation>
		<dc:coverage></dc:coverage>
		<dc:rights></dc:rights>
		**<dc:language>en-US-POSIX</dc:language>**
		<dc:language>pt-BR</dc:language>
		<dc:identifier id="bookid">xxx</dc:identifier>
	</metadata>
```

`<dc:language>en-US-POSIX</dc:language>` doesn't have a valid value and epubcheck ignores that.

epubcheck output:
```
java -jar epubcheck.jar xxx.epub 
Validating using EPUB version 2.0.1 rules.
No errors or warnings detected.
epubcheck completed
```

kalaspuffar:
With this implementation, I try to fix the issue of #702.

This current fix will generate an error if the language prefix is longer than 3 characters.
The specification says that you could have pre-registered language subtags or could use 4 letter codes but these are reserved for future use and currently not used to my knowledge.

If this test passes it will check if this language has a variant specified and if so it will generate a warning informing the user of this discrepancy.

I hope this fix will start discussion towards a complete solution.

### DIFF
--- a/src/main/java/com/adobe/epubcheck/messages/MessageDictionary.java
+++ b/src/main/java/com/adobe/epubcheck/messages/MessageDictionary.java
@@ -283,7 +283,7 @@ public class MessageDictionary
       map.put(MessageId.OPF_082, Severity.ERROR);
       map.put(MessageId.OPF_083, Severity.ERROR);
       map.put(MessageId.OPF_084, Severity.ERROR);
-      map.put(MessageId.OPF_085, Severity.WARNING);
+      map.put(MessageId.OPF_085, Severity.INFO);
       map.put(MessageId.OPF_086, Severity.ERROR);
 
       // PKG
--- a/src/main/java/com/adobe/epubcheck/messages/MessageDictionary.java
+++ b/src/main/java/com/adobe/epubcheck/messages/MessageDictionary.java
@@ -283,6 +283,8 @@ public class MessageDictionary
       map.put(MessageId.OPF_082, Severity.ERROR);
       map.put(MessageId.OPF_083, Severity.ERROR);
       map.put(MessageId.OPF_084, Severity.ERROR);
+      map.put(MessageId.OPF_085, Severity.WARNING);
+      map.put(MessageId.OPF_086, Severity.ERROR);
 
       // PKG
       map.put(MessageId.PKG_001, Severity.WARNING);
--- a/src/main/java/com/adobe/epubcheck/opf/OPFChecker30.java
+++ b/src/main/java/com/adobe/epubcheck/opf/OPFChecker30.java
@@ -26,6 +26,7 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.HashMap;
 import java.util.Iterator;
+import java.util.Locale;
 import java.util.Set;
 
 import com.adobe.epubcheck.api.EPUBLocation;
@@ -89,6 +90,7 @@ public class OPFChecker30 extends OPFChecker implements DocumentValidator
     checkSemantics();
     checkNav();
     checkSpecifics();
+    checkLanguage();
   }
 
   @Override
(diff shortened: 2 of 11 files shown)


tofi86:
Hej Daniel, thanks for your contribution! I'll try to review on the weekend!
Best, Tobias

tofi86:
@kalaspuffar Daniel, is it this statement which makes the PullRequest a WorkInProgress at the moment?

> This current fix will generate an error if the language prefix is longer than 3 characters.
The specification says that you could have pre-registered language subtags or could use 4 letter codes but these are reserved for future use and currently not used to my knowledge.

If so: @murata0204 and @mattgarrish as you've been involved in the discussion about the topic in the original issue, could you probably answer Daniel's question?

src/main/java/com/adobe/epubcheck/messages/MessageDictionary.java
```diff
@@ -282,6 +282,8 @@ public Message getMessage(MessageId id)
       map.put(MessageId.OPF_082, Severity.ERROR);
       map.put(MessageId.OPF_083, Severity.ERROR);
       map.put(MessageId.OPF_084, Severity.ERROR);
+      map.put(MessageId.OPF_085, Severity.WARNING);
```
rdeltour: I don't think EpubCheck can issue a WARNING for language tag variants, as there are legitimate uses for the variant tag. See for instance [W3C’s explainer on the language tags](https://www.w3.org/International/articles/language-tags/#variants).

At best this can be a HINT, but then I'm not quite sure it's worth implementing if it's just informative?

src/main/java/com/adobe/epubcheck/opf/OPFChecker30.java
```diff
@@ -395,6 +397,26 @@ private void checkPagination()
...
+  private void checkLanguage() {
+    Set<Metadata> dcLanguageMetas = ((OPFHandler30) opfHandler).getMetadata()
+            .getPrimary(DCMESVocab.VOCAB.get(DCMESVocab.PROPERTIES.LANGUAGE));
+    if (!dcLanguageMetas.isEmpty())
+    {
+      for(Metadata metadata : dcLanguageMetas)
+      {
+        Locale l = Locale.forLanguageTag(metadata.getValue());
+        if (l == null || l.getLanguage().length() > 3) {
+          report.message(MessageId.OPF_086, EPUBLocation.create(path), metadata.getValue());
```
rdeltour: OK. 
A more thorough validation would be to use [IANA’s language tag registry](https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry), but this is probably overkill?

kalaspuffar:
Hi @rdeltour 

I've changed the warning for variants to info.

You mention that this check should be done in the schema already. Didn't see one of those checks and my implementation only use Locale.forLanguageTag from Java 1.7 to validate that you use a valid format.

And if you have a variant defined I will inform you about that too. 

Maybe this could be done "easier" or more correct by a validation rule in the schema?

rdeltour:
Closing this PR for now. A more complete check (based on the IANA registry) may be implemented in the future, depending on the outcome of the spec discussion at w3c/epub-specs#1509

PR #809, opened 2017-11-23
[fix #289] Check if a file contains a file:// link

### LINKED ISSUE #289: URL in href attribute started by file://
_From [vgros00@gmail.com](https://code.google.com/u/vgros00@gmail.com/) on July 04, 2013 12:32:04_

PROBLEM: URL like `file:///C:/aFolder/afile.pdf` should not be used in EPUB file.

METHOD: For example:

```
<a class="external" href="file:///C:/aFolder/afile.pdf">
```

RESULT: No error or warning.

EXPECTED: At least a warning.

PLATFORM: Windows 7 pro.

ADDITIONAL INFO: See attached files.

**Attachment:** [SimpleSample2_file-in-href.epub SimpleSample3_file-in-href.epub](http://code.google.com/p/epubcheck/issues/detail?id=289)

_Original issue: http://code.google.com/p/epubcheck/issues/detail?id=289_

kalaspuffar:
Implemented a fix for issue #289.

This fix includes some extra code to validate info messages in tests.
It also abstracts some code from EntitySearch and implements a new FileLinkSearch.

The solution works but I'm open to discussion.


rdeltour:
Thanks for the fix @kalaspuffar! 👍 

One comment however: we should ideally keep changes to the `com.adobe.epubcheck.ctc` minimal, as in the longer term this more recent package should be fully reviewed and possibly integrated at other place in EpubCheck (as hinted in the [work plan in the wiki](https://github.com/IDPF/epubcheck/wiki/WorkPlan#removal--integration-of-the-nook-contributions)).

There's no way you could have known this, so it's absolutely fine! but for future changes it's always better to try and see if it can be done in other packages first 😄

PR #811, opened 2017-11-26
[fix #804] improved messages for OPF-058 and OPF-059

### LINKED ISSUE #804: Improve OPF-058/059 usage messages
The [OPF-058](https://github.com/IDPF/epubcheck/blob/master/src/main/resources/com/adobe/epubcheck/messages/MessageBundle.properties#L217) and [OPF-059](https://github.com/IDPF/epubcheck/blob/master/src/main/resources/com/adobe/epubcheck/messages/MessageBundle.properties#L219) usage messages state that `"Spine item has no TOC/NCX entry reference."` without mentioning _which_ item the message actually refers to.

However, that information is available when the message is reported in [`EpubNavCheck.java:196`](https://github.com/IDPF/epubcheck/blob/master/src/main/java/com/adobe/epubcheck/ctc/EpubNavCheck.java#L196) and [`EpubNCXCheck.java:176`](https://github.com/IDPF/epubcheck/blob/master/src/main/java/com/adobe/epubcheck/ctc/EpubNCXCheck.java#L176) as the location’s `path` property. Looking at the report message formatter (for terminal use in my case, [`DefaultReportImpl.java:90`](https://github.com/IDPF/epubcheck/blob/master/src/main/java/com/adobe/epubcheck/util/DefaultReportImpl.java#L90)) I see that the location’s `path` property (i.e. `location.getPath()`) isn’t actually being added to the usage message. Adding it breaks a bunch of test cases so I did not pursue this matter further for now.

I think it would be quite useful if the usage messages contained the above path information.

tofi86:
improved messages for OPF-058 and OPF-059

fixes #804


tofi86:
man, the JSON tests are killing me... :-/ cleaned up all the tests in `css` but still 40 to go... :-(

tofi86:
Fixed the failing tests by copying the actual results with a bash script as the expected results.
No more cleanup. But to be honest, this is a mess! 😲  OPF_058 and OPF_059 are everywhere where they aren't supposed to be...

tofi86:
For further reference, this was the bash script, executed from epubcheck root dir:

```bash
#!/bin/bash

for actualResult in target/test-classes/com/adobe/epubcheck/test/*/*_actual_results.*; do
  expectedSource=${actualResult/target\/test-classes/src\/test\/resources}
  expectedSource=${expectedSource/actual_results/expected_results}

  if grep -Eq "OPF-05(8|9)" $actualResult; then
    echo $actualResult
    cp $actualResult $expectedSource
    echo "=> $expectedSource"
  fi
done
```

PR #812, opened 2017-11-28
Added warning about reimplementation to classes in package ctc.

kalaspuffar:
Hi @rdeltour 

I'm not sure if this is in order but I thought it might be a good thing to send a strong message that these classes in this package will be replaced. The comment "the grouped most of their changes in a single package" isn't helpful if not mentioning which package :)

https://github.com/IDPF/epubcheck/pull/809#issuecomment-347363673

Another "huge" undertaking might be to at lease comment each class with their purpose so you could generate a readable java doc to navigate the code easier but I understand that this is something that needs to be done over time.

best regards
Daniel

### DIFF
--- a/src/main/java/com/adobe/epubcheck/ctc/CheckManager.java
+++ b/src/main/java/com/adobe/epubcheck/ctc/CheckManager.java
@@ -7,11 +7,12 @@ import com.adobe.epubcheck.ctc.ContentValidator.ValidationType;
 import com.adobe.epubcheck.ctc.epubpackage.EpubPackage;
 import com.adobe.epubcheck.util.EPUBVersion;
 
-/*  ===  WARNING  ==========================================
- *  This class is scheduled to be refactored and integrated
- *  in another package.
- *  Please keep changes minimal (bug fixes only) until then.
- *  ========================================================
+/**
+ *  ===  WARNING  ==========================================<br/>
+ *  This class is scheduled to be refactored and integrated<br/>
+ *  in another package.<br/>
+ *  Please keep changes minimal (bug fixes only) until then.<br/>
+ *  ========================================================<br/>
  */
 public class CheckManager
 {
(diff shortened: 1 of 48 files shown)


rdeltour:
The thing is this package is currently in use and I'm afraid that it's too early to formerly "deprecate" it, or that it wouldn't send the right message.  Maybe we should add a warning comment at the top of each of its class file instead? WDYT @kalaspuffar?

> The comment "the grouped most of their changes in a single package" isn't helpful if not mentioning which package :)

Good point, I fixed the wiki to name the package explicitly 😄 

> Another "huge" undertaking might be to at lease comment each class with their purpose so you could generate a readable java doc to navigate the code easier but I understand that this is something that needs to be done over time.

👍 
It's kinda hidden in the draft work plan as "All public API should be properly documented", but I agree that it shouldn't be limited to public API.
I added a new ["Code documentation" section](https://github.com/IDPF/epubcheck/wiki/WorkPlan#code-documentation) to the work plan.

kalaspuffar:
Hi @rdeltour 

Well, a deprecation only means we shouldn't build or rely on it. Many things in java have been deprecated for years, now they finally remove some of them.

Then again adding a warning on each class could be done as well. Writing something like

```
******* WARNING ********
This class was a part of a large contribution by Barnes & Nobel back in 2013.
This class is scheduled to be reimplemented.
```

Not sure if we need to add the original author to the class as well? 
Author information is only interesting if you need to contact someone for development support and it's hard to maintain these tags if put all over the code.

Are all the classes in this package of the original package or is there some that shouldn't have the warning?

best regards
Daniel

rdeltour:
> Well, a deprecation only means we shouldn't build or rely on it. Many things in java have been deprecated for years, now they finally remove some of them.

Right, but I would hesitate to deprecate something when we don't have an alternative yet.

> Then again adding a warning on each class could be done as well. 

Yes, I think that's preferable for now. Let's use this text:

```java
/*  ===  WARNING  ==========================================
 *  This class is scheduled to be refactored and integrated
 *  in another package.
 *  Please keep changes minimal (bug fixes only) until then.
 *  ========================================================
 */
```

> Not sure if we need to add the original author to the class as well?

No, I don't think it's necessary.

> Are all the classes in this package of the original package or is there some that shouldn't have the warning?

I think it applies to all the classes in the package.

kalaspuffar:
Hi @rdeltour 

I've added the warning to all classes in the package.

Best regards
Daniel

tofi86:
Merged! Thanks, Daniel! 👏

PR #817, opened 2017-12-04
[Fix #813] Cleaning up OPF-058 and OPF-059 from tests

### LINKED ISSUE #813: Clean tests from OPF-058 and OPF-059 messages
`OPF-058` and `OPF-059` messages are scattered all along the tests in `src/test/resources/com/adobe/epubcheck/test/` and should be cleaned up by fixing the `toc.ncx` and `toc.xhtml` files in the particular tests. The messages appear because the guys from B&N/Nook built all theier tests on 'invalid' EPUBs. `OPF-058` and `OPF-059` are not explicitly checked in the tests and therefore have no place there.

Two new dedicated testcases for `OPF-058` and `OPF-059` need to be created based on our new minimal epub testdata in `src/test/resources/minimal-epub/` and a dedicated JUnit test needs to be set up.

I cleaned up a first batch of tests with commit 7d7365a8c9b93d7a86634fc99c6ab3a7c46a3b26 in PR #811 but didn't have the time to finish the rest of 40 other tests. You can identify them by looking at commit 11cb2452d8029bac7d31c136128902202e3a7fb2.

### Tasks
- [ ] Clean the tests
- [ ] Fix the *expected* JSON/XML files
- [ ] Create minimal epub tests
- [ ] Create JUnit tests for both new tests

kalaspuffar:
I started by cleaning up the excessive testing because these were the largest and required the most work.


tofi86:
Awesome, thanks!
refs #813

kalaspuffar:
Hi @tofi86 

When running the tests I came across the test
com/adobe/epubcheck/test/package/interesting_paths_epub3_expected_results.json

Not really sure what it should test. It tests so many things and throws so many exceptions. And when I was done with it, it threw OPF-010 which has been commented as had to reproduce.

My approach was to go with the files inside of the directory. Another approach could be to actually use the phi character in the filename and get another set of error codes that you shouldn't use greek chars in filenames. Hard to figure out what was the actual test case and what was weird side effects.

Best regards
Daniel

kalaspuffar:
Hi @tofi86 

Another test case that was a bit weird was
com/adobe/epubcheck/test/opf/Mismatched_mimetypes_epub2.epub

Which generated 2 new errors:
RSC-010 : Reference to non-standard resource type found.
PKG-021 : Corrupted image file encountered.

Best regards
Daniel

tofi86:
Hi Daniel, thanks for your work on this!

Regarding the testcase `Mismatched_mimetypes_epub2`:
PKG-021 has already been in the old results: https://github.com/IDPF/epubcheck/pull/817/files#diff-7910489acff575ddcbc72a9498973debL372
Whereas RSC-010 is indeed new because the NCX now contains a reference to a spine item with an invalid/unknown mimetype: https://github.com/kalaspuffar/epubcheck/blob/c24b20f24bccac7bd791f0f3ff8b93af8cb897cd/src/test/resources/com/adobe/epubcheck/test/opf/Mismatched_mimetypes_epub2/OPS/content.opf#L11
```
<item id="page02" href="page02.xhtml" media-type="application/xhtml"/>
```
The mimetype is wrong, it should be `application/xhtml+xml`. If you fix that, the error will be gone.

tofi86:
Regarding the `interesting_paths_epub3` testcase I think we should retain the cyrillic phi character `Ф` in the filename and adapt the tests accordingly. Alle filenames in this testcase contain "special characters" so I assume the original tests should check for those...

kalaspuffar:
Hi @tofi86 

Just to check the ```Mismatched_mimetypes_epub2``` has the name Mismatched mimetypes so I didn't change the test data. You could interpret ```application/xhtml``` as mismatched.

But we should change the data and reduce the numbers of errors?

Best regards
Daniel

tofi86:
hej Daniel, absolutely right, I should have taken more time to look at this. the testcase is all about mismatching mimetypes so we don't fix that and I suggest to leave the tests as they are.

Fixing the TOC lead to the new error `RSC-010` because the TOC only allows references to certain core-media types and `application/xhtml` isn't one of those.
I would leave it as it is because the `Mismatched_mimetypes_epubX` tests seem to be the only ones throwing this error code.

tofi86:
Merged! Thanks, Daniel! 🎉

PR #818, opened 2017-12-05
Path resolution fix

### LINKED ISSUE #813: Clean tests from OPF-058 and OPF-059 messages

kalaspuffar:
When I went through the test in order to fix #813 I couldn't fix one test case where the path of the package was different from the NCX file.

I realized that this was a bug and I separated this cleanup into this pull request fixing the bug.

### DIFF
--- a/src/main/java/com/adobe/epubcheck/ctc/EpubNCXCheck.java
+++ b/src/main/java/com/adobe/epubcheck/ctc/EpubNCXCheck.java
@@ -169,7 +169,7 @@ public class EpubNCXCheck implements DocumentValidator
           if (mi != null)
           {
             String path = mi.getHref();
-            path = PathUtil.resolveRelativeReference(navDocEntry, path,  null);
+            path = PathUtil.resolveRelativeReference(epack.getPackageMainFile(), path,  null);
 
             if (path != null && !path.equals(tocFileName) && !path.equals(navDocEntry) && !tocLinkSet.contains(path))
             {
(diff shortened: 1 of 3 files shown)


kalaspuffar:
Hi @tofi86 

Just to clarify this issue I want to present this piece of code.

```
PackageManifest manifest = epack.getManifest();
ManifestItem mi = manifest.getItem(si.getIdref());
String path = mi.getHref();
            path = PathUtil.resolveRelativeReference(epack.getPackageMainFile(), path,  null);
```

We get a package manifest, we get one item of this manifest, the Href of that item and lastly resolve the relative reference of that item.

To use the navDocEntry to figure out the reference will only work when the package file and the ncx file is in the same directory. This test case test exactly this and this haven't been found because the error OPF-059 is the outcome of such errors.

The directory structure in this test:
```
OPS/toc.ncx
OPS/toc.xhtml
content.opf
```

Maybe we should have a test like this as well?
```
```
toc.ncx
toc.xhtml
OPS/content.opf
```

Best regards
Daniel

Issue #743, opened 2017-01-12
Add a --version switch


tofi86:
The `epubcheck.jar` should have a `--version` switch to display the version information only.

tofi86:
hey @rdeltour, we already parse the argument string for a `--version` string in https://github.com/IDPF/epubcheck/blob/master/src/main/java/com/adobe/epubcheck/tool/EpubChecker.java#L541

Seems this was introduced as an alternative to the `-v` switch when using in conjunction with `--mode` / `-m`.

However, it's not documented anywhere. Not in commandline help `--help` nor in the wiki: https://github.com/IDPF/epubcheck/wiki/Running

What do you think, shall we just drop support for alternates `-version` and `--version` and make `--version` print the EpubCheck version number?

tofi86:
Postponed to a later version, as this is causing conflicts with the current commandline arguments

kamorrissey:
I'd like to get my feet wet as a contributor, and this seemed like a modest item. If I understand correctly, the "ready for implementation" tag makes it suitable to work on. Is this right?

tofi86:
Glad to see new volunteers, @kamorrissey! Welcome!

If we can agree on having two different meanings for the `--version` switch, this issue is ready to go!
I can live with it tbh.

rdeltour:
> Glad to see new volunteers, @kamorrissey! Welcome!

+1, hi and welcome @kamorrissey 👋

> If we can agree on having two different meanings for the --version switch, this issue is ready to go!
> I can live with it tbh.

It seems the current `--version` (or `-v`) option is used to declare the EPUB spec version against which to check when validating single documents (with the `--mode` option), since in that case we can’t get the EPUB version from the OPF.

I agree that it can be pretty confusing with `--version` typically being used in a CLI to display the tool’s version.

We’d have a couple options:
1. remove `--version` from its previous usage, and use it to display the tool’s version
2. rename the `-v`|`--version` option to something else (e.g. `-s`|`--spec-version`, or `-t`|`--target-version`)
3. keep the current use of the `--version` option, and make it display the tool’s version number only when the mode `--mode` is *not* specified

I don’t have a strong opinion; I don’t know if this option is heavily used out there. Option 3 is the less disruptive, but the most confusing.

kamorrissey:
I've been considering it. I'm in favor of option 1, since the --help message does not mention --version. I already have a proposed implementation that I'm going to submit in a pull request.

rdeltour:
[...]

PR #820, opened 2017-12-14
First implementation of a simple BDD test.

kalaspuffar:
Hi again.

After yesterdays meeting I thought I set up a simple cucumber BDD test so you can see how such a test could be implemented.

The nice thing with this framework is that it runs along with our current maven tests. And the test writer could just write a feature file in a new branch and then the implementer would fetch that branch run the test and create the implementation.

One thing to note is that cucumber only works with jUnit >= 4.11. And in that version, the is() function is removed and replaced with instanceOf. There is a lot of other functions deprecated but still supported.

### DIFF
--- a/src/test/resources/com/adobe/epubcheck/bdd/general.feature
+++ b/src/test/resources/com/adobe/epubcheck/bdd/general.feature
@@ -1,0 +1,7 @@
+Feature: General
+  As a user
+  I want the application to display information clearly
+
+  Scenario: [MUST] Visible version number
+    When I run the application with "--help"
+    Then The version number should be clearly visible on the first line
--- a/pom.xml
+++ b/pom.xml
@@ -193,12 +193,6 @@
             <version>1.3</version>
         </dependency>
         <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
-            <version>4.10</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
             <version>14.0.1</version>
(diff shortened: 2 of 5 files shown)


kalaspuffar:
A thought could be to use this to implement #743 later. 

Can't see that --version really collides with anything because of the slight that it's an undocumented feature. Using undocumented features should be hazardous anyway.

src/test/resources/com/adobe/epubcheck/bdd/general.feature
```diff
@@ -0,0 +1,7 @@
+Feature: General
+  As a user
+  I want the application to display information clearly
+
+  Scenario: [MUST] Visible version number
+    When I run the application with "--help"
```
BigBlueHat: Should this be `--version` instead of `--help`? The test says "flag version."

BigBlueHat:
Thanks for doing this @kalaspuffar! This seems much clearer (granted, it's one chunk of code to test one simple feature 😄). I do like how it reads to be sure.

tofi86:
That looks very interesting indeed! Thanks for doing a first implementation, Daniel! 👍

src/test/resources/com/adobe/epubcheck/bdd/general.feature
kalaspuffar: Hi @BigBlueHat

You're right. Today that functionallity is missing though. As mentioned in the pull request it can be changed when we decide how.to do with the --version.

Today the --version is undocumented feature to specify file version for some modes. We only documented -v so changing should not break anything.

rdeltour:
Closing this PR. The BDD approach was successfully adopted for the cleanup process. A whole range of Cucumber steps used in EPUBCheck tests are currently documented in the [Test Suite](https://github.com/w3c/epubcheck/wiki/TestSuite) wiki page.

Thanks for the exploratory work @kalaspuffar !

PR #854, opened 2018-08-07
[fix #853] add UUID validation

### LINKED ISSUE #853: UUID <dc:identifier> should be validated
I had a couple of customer requests lately about invalid UUID's like `urn:uuid:978...` where people placed a ISBN as a UUID. Adobe Content Server is checking UUID correctness and therefore this error is raised in multiple epub delivery pipelines.

In previous versions of EpubCheck, this check was implemented. See #109. Current versions do not report an invalid UUID.

I'm going to submit a PR to get this fixed again!

tofi86:
see #853 for details


tofi86:
The UUID function seems to have issues with JDK 9/10 👿 . Need to investigate.

PR #911, opened 2018-11-24
Allow the configuration of EPUBCheck’s Locale

### LINKED ISSUE #498: Using i18n of releases/builds
How can the i18n of releases/builds be used? Neither the readme nor the Wiki nor the --help command line argument seem to give a hint how to switch to another language.

rdeltour:
This PR completes and replaces #650.

The locale used by EPUBCheck can now be configured:

- with a `--locale` option on the command line
- with the `setLocale(Locale locale)` method on the `EpubCheck` class
- when no locale is provided, the library will fall back to the default
  JVM Locale (i.e. EPUBCheck’s behavior doesn’t change)

This PR introduces the following changes:

- Refactored `com.adobe.epubcheck.util.Messages`:
  - new static `getInstance(…)` methods to get the `Messages` instance
    for the given (optional) locale and class name (used to resolve
    the name of the underlying `ResourceBundle`).
  - the `Messages` instances are cached in a table, keyed with locale
    codes and resource bundle names.
- Refactored the `com.adobe.epubcheck.message` package:
  - `MessageDictionary` is now an interface; it's functionality is split
    in several classes to handle default messages, loading of override
    files, overridden messages, dumping message files, and severities.
  - messages and severities are cached.
- Refactored the `org.idpf.epubcheck.util.css` package, to make it
  locale-aware:
    - a `Locale` is given as an argument to the `CSSParser` constructor
    - the package `Messages` class is removed, replaced by the main
      `com.adobe.epubcheck.util.Messages`
- New `LocalizedReport` interface, extending the `Report` interface with
  locale getter/setter methods, representing a localizable report.
  - the existing `Report` interface is kept unchanged, for backwards
    compatibility
  - `MasterReport` implements this new `LocalizedReport` interface
  - the localization code checks if the underlying report is an instance
    of `LocalizedReport` before setting a new locale.
- New `LocaleHolder` facility to statically store the "current" locale
  in a thread-local variable. This is used by libraries that cannot be
  configured dynamically with a new locale (like Jing).
- Monkey-patch Jing's `Localizer` class to load Jing's resource bundles
  for the locale exposed in the `LocaleHolder` utility.
- Expose the current locale in the `ValidationContext` class
- Update the java sources compatibility version to 1.7
- New locale-related tests

Limitations:

- Jing's localization relies on the locale information exposed in a
static thread-local variable (in the `LocaleHolder` class). Any new
implementations of the `LocalizedReport` must make sure that the locale
stored in the `LocaleHolder` is updated when the locale is changed or
initialized.
- The validation messages defined in the schematron schemas are still
not localized.

Thanks to @matthew-macgregor for the original PR contribution (#650)!

Closes #650
Fixes #498


rdeltour:
thanks for the testing @tofi86! 

> --locale has currently no effect on the --help message, but I think this is acceptable for the moment.

Mmm, weird, it does work for me… note that currently the `--locale` option has to be set **before** the `--help` option to be taken into account, maybe that's why?

tofi86:
> Mmm, weird, it does work for me…

Sorry, I tried `--locale=fr` instead of `--locale fr` yesterday. My fault.

Works fine!

PR #918, opened 2018-11-27
Some OSGi related changes

bertfrees:
- The version of Saxon needs to be restricted to a more specific range because Saxon doesn't do semantic versioning.
- Update to more recent version of maven-bundle-plugin to solve a problem when using two versions of jing (20120724.0.0 + a newer one) in the same OSGi application.


bertfrees:
Note that it would also be possible to update to a more recent version of Jing, but this is not a requirement for us.

PR #927, opened 2018-12-17
Updated package schemas to check new 3.2 rules

### LINKED ISSUE #876: Add checks for authority and term with dc:subject
If the authority property is attached (via refines) to a dc:subject, then a term property also has to be attached. Likewise, if a term is attached, there must be a corresponding authority attached (i.e., they must always appear together).

Reference: https://w3c.github.io/publ-epub-revision/epub32/spec/epub-packages.html#sec-opf-dcsubject

### LINKED ISSUE #877: Use of display-seq meta property is deprecated
Epubcheck should emit a warning if the property is used in the meta property attribute.

Reference: https://w3c.github.io/publ-epub-revision/epub32/spec/epub-packages.html#sec-display-seq

### LINKED ISSUE #878: Use of meta-auth meta property is deprecated
Epubcheck should emit a warning if the property is used in the meta property attribute.

Reference: https://w3c.github.io/publ-epub-revision/epub32/spec/epub-packages.html#sec-meta-auth

### LINKED ISSUE #879: Use of rendition:spread value "portrait" is deprecated
Epubcheck should emit a warning that the rendition:spead "portrait" value is deprecated when used in the package document metadata. The "both" value is recommended instead.

The spread-portrait spine override on itemref elements is also deprecated.

Reference: https://w3c.github.io/publ-epub-revision/epub32/spec/epub-packages.html#spread-values

### LINKED ISSUE #880: Use of rendition:viewport is deprecated
Epubcheck should emit a warning if rendition:viewport is used in the package document metadata.

Reference: https://w3c.github.io/publ-epub-revision/epub32/spec/epub-packages.html#viewport

### LINKED ISSUE #881: Use of bindings is deprecated
Epubcheck should emit a warning that use of the bindings element in the package document is deprecated.

Reference: https://w3c.github.io/publ-epub-revision/epub32/spec/epub-packages.html#sec-opf-bindings

### LINKED ISSUE #882: Duplicate itemref references not allowed
It is an error for two spine itemrefs to reference the same manifest item.

Reference: https://w3c.github.io/publ-epub-revision/epub32/spec/epub-packages.html#attrdef-itemref-idref

### LINKED ISSUE #883: Add properties attribute to link element
The link element now allows properties attribute. The value(s) of the attribute must be prefixed unless they are defined in the [link properties vocabulary](https://idpf.github.io/epub-vocabs/package/link/#sec-link-properties).

Reference: https://w3c.github.io/publ-epub-revision/epub32/spec/epub-packages.html#sec-link-elem

mattgarrish:
Updates to the package schema for the following issues:

- partially completes #883 - only adds the properties attribute to link in the rnc, doesn't check values
- fixes #882 - reports if a preceding itemref has a matching idref
- fixes #881 - reports presence of bindings
- fixes #880 - reports presence of rendition:viewport property in metadata
- fixes #879 - reports rendition:spread=portrait in metadata and rendition:spread-portrait on an itemref
- fixes #878 - reports presence of meta-auth property in metadata
- fixes #877 - reports presence of display-seq in metadata
- fixes #876 - reports if term/authority are not used together, or if multiple instances are attached to a single subject


danielweck:
Related PR:
https://github.com/w3c/epubcheck/pull/934

src/main/resources/com/adobe/epubcheck/schema/30/package-30.sch
```diff
@@ -295,5 +295,62 @@
...
+	
+	<pattern id="opf.meta.spread.portrait.deprecated">
+		<rule context="opf:metadata/opf:meta[@property='rendition:spread']">
+			<report test=". = 'portrait'">WARNING: Use of the rendition:spread value 'portrait' is deprecated in favour of the value 'both'</report>
+		</rule>
+	</pattern>
+	
+	<pattern id="opf.itemref.spread.portrait.deprecated">
+		<rule context="opf:spine/opf:itemref[@properties]">
+			<report test="(some $token in tokenize(@properties,' ') satisfies (normalize-space($token) eq 'rendition:spread-portrait'))">WARNING: Use of the 'rendition:spread-portrait' spine override is deprecated in favour of 'rendition:spread-both'</report>
```
rdeltour: the XPath could be simplified to `tokenize(@properties,'\s+')='rendition:spread-portrait'`

src/test/java/com/adobe/epubcheck/api/Epub30CheckExpandedTest.java
```diff
@@ -133,19 +133,22 @@ public void testValidateEPUBPLoremForeign()
   @Test
   public void testValidateEPUBPLoremObjectFallbacks()
   {
+    Collections.addAll(expectedWarnings, MessageId.RSC_017);
```
rdeltour: Ideally for all these tests where the content is expected to be valid, the data files should be fixed instead of adding an expected warning.
Otherwise a long list of warnings and errors is gradually added to each an every test, and it gets harder to know what the test was actually suppose to focus on.

src/test/resources/30/single/opf/invalid/prefixes-001.opf
```diff
@@ -8,8 +8,7 @@
         <dc:language>la</dc:language>
         <dc:date>2011-09-01</dc:date>
         <meta property="dcterms:modified">2011-09-01T17:18:00Z</meta>
-		<meta property="meta-auth">Auth</meta>
-		<meta property="media:test meta-auth">test</meta>
```
rdeltour: this test case was used to verify that multiple property values were allowed and detected; we should replace `meta-auth` by another valid property value.

src/test/resources/30/single/opf/invalid/prefixes-002.opf
```diff
@@ -8,7 +8,6 @@
         <dc:language>la</dc:language>
         <dc:date>2011-09-01</dc:date>
         <meta property="dcterms:modified">2011-09-01T17:18:00Z</meta>
-		<meta property="meta-auth">Auth</meta>
```
rdeltour: likewise, this test case may be used to check that properties in the default Vocabulary were recognized. To be verified, and if yes `meta-auth` should be replaced.

src/main/resources/com/adobe/epubcheck/schema/30/package-30.sch
```diff
@@ -295,5 +295,62 @@
     </pattern>
     
     <include href="./mod/id-unique.sch"/>
-
-</schema>
+	
+	<!-- EPUB 3.2 New Checks -->
```
rdeltour: These changes look good!
However before merging we should test all these new warnings, each in its own test case…

src/test/java/com/adobe/epubcheck/api/Epub30CheckExpandedTest.java
mattgarrish: Right, but there were a number of "valid" files that depended on deprecated features. That's typically where I added warnings (e.g., I believe the one referenced here was using bindings on the object's media type).

src/test/resources/30/single/opf/invalid/prefixes-001.opf
mattgarrish: Is there documentation of what these tests are for? I assumed by being an "invalid" test with the name "prefixes", it was checking for the undeclared "media:" prefix.

mattgarrish: Isn't this the only relevant error in the file:

mattgarrish: But going to the source again, isn't this the duplication error being checked:

```
<link rel="marc21xml-record foaf:ddd foaf:ddd" href="pub/meta/nor-wood-marc21.xml"/>
```

src/main/resources/com/adobe/epubcheck/schema/30/package-30.sch
```diff
@@ -295,5 +295,62 @@
...
+	<pattern id="opf.spine.duplicate.refs">
+		<rule context="opf:itemref[@idref = preceding-sibling::opf:itemref/@idref]">
+			<report test=".">Itemref refers to the same manifest entry as a previous itemref</report>
+		</rule>
+	</pattern>
+	
+	<pattern id="opf.subject.authority-term">
+		<rule context="opf:metadata/dc:subject">
+			<let name="id" value="./@id"/>
+			<let name="authority" value="//opf:meta[@property='authority'][substring(@refines, 2) = $id]"/>
```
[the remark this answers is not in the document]
mattgarrish: Yikes, yes. Looking through the old tests, there are lots that won't check what they're supposed to if there is any whitespace around them. I'm going through and updating them all now to add normalize-space where needed.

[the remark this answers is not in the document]
mattgarrish: It's actually a problem all over with attribute tests. XML processors will normalize whitespace in attributes, but there are lots of tests where we're not taking that into account. For example, `id="    foo    "`  is valid, but `unique-identifier="foo"` won't properly test it, and vice versa. In some cases, like the previous, it won't matter because an exact match is required, but overall the tests are a combination of being too brittle and too strict.

src/test/resources/30/single/opf/invalid/prefixes-001.opf
rdeltour: > Is there documentation of what these tests are for?

No. We have to "guess" from the test file’s or test method’s names or contents. Which is another reasons why test maintenance is difficult 🤷‍♂️

> I assumed by being an "invalid" test with the name "prefixes", it was checking for an undeclared prefix.

Possibly. I'll go over all these and try to figure this out!

PR #928, opened 2018-12-17
Add the Compression element

### LINKED ISSUE #904: Add support for Compression element
Add support for the Compression element to validation of the encryption.xml file.

Reference: https://w3c.github.io/publ-epub-revision/epub32/spec/epub-ocf.html#elemdef-enc-Compression

mattgarrish:
This PR adds the Compression element that was introduced in the 3.1 schema. Fixes #904

### DIFF
--- a/src/test/java/com/adobe/epubcheck/ocf/OCFCheckerTest.java
+++ b/src/test/java/com/adobe/epubcheck/ocf/OCFCheckerTest.java
@@ -306,6 +306,20 @@ public class OCFCheckerTest
   }
 
   @Test
+  public void testValidCompression()
+  {
+    ValidationReport testReport = testOcfPackage("/30/expanded/valid/ocf-compression/",
+        EPUBVersion.VERSION_3);
+
+    if (0 != testReport.getErrorCount() || 0 != testReport.getWarningCount())
+    {
+      outWriter.println(testReport);
+    }
+    assertEquals(0, testReport.getErrorCount());
+    assertEquals(0, testReport.getWarningCount());
+  }
+
+  @Test
   public void testInvalidLoremForeign30()
   {
     ValidationReport testReport = testOcfPackage("/30/expanded/invalid/lorem-foreign/",
@@ -383,5 +397,28 @@ public class OCFCheckerTest
     assertEquals(0, testReport.getWarningCount());
 
     assertTrue(testReport.hasInfoMessage(VERSION_STRING));
+  }
+
+  @Test
+  public void testInvalidCompressionMethod()
+  {
+    ValidationReport testReport = testOcfPackage("/30/expanded/invalid/ocf-compression/",
+        EPUBVersion.VERSION_3);
+
+    if (2 != testReport.getErrorCount() || 0 != testReport.getWarningCount())
+    {
+      outWriter.println(testReport);
+    }
+    List<MessageId> errors = new ArrayList<MessageId>();
+    Collections.addAll(errors, MessageId.RSC_005, MessageId.RSC_005);
+    assertEquals(errors, testReport.getErrorIds());
+    assertEquals(0, testReport.getWarningCount());
+    if (testReport.errorList.size() >= 2)
+    {
+      assertTrue(testReport.errorList.get(0).message
+          .contains("value of attribute \"Method\" is invalid; must be equal to \"0\" or \"8\""));
+      assertTrue(testReport.errorList.get(1).message
+          .contains("value of attribute \"OriginalLength\" is invalid; must be an integer"));
+    }
   }
 }
--- a/src/main/resources/com/adobe/epubcheck/schema/30/ocf-encryption-30.rnc
+++ b/src/main/resources/com/adobe/epubcheck/schema/30/ocf-encryption-30.rnc
@@ -1,12 +1,30 @@
-   default namespace = "urn:oasis:names:tc:opendocument:xmlns:container"
-   start = element encryption {
-     grammar {
-       include "mod/security/Lenient-Encryption11-ghc.rnc"  {
-         start = xenc_EncryptedData | xenc_EncryptedKey
-       }
-     xenc_EncryptionMethodOtherParams |=
-        attribute Algorithm { xsd:anyURI "http://www.idpf.org/2008/embedding" }
-     xenc_EncryptionAlgorithms |=
-        xsd:anyURI "http://www.idpf.org/2008/embedding"
-     }+
-   }
+default namespace = "urn:oasis:names:tc:opendocument:xmlns:container"
+namespace epub_xenc_compress = "http://www.idpf.org/2016/encryption#compression"
+namespace xenc = "http://www.w3.org/2001/04/xmlenc#"
+
+start =
+	element encryption {
+		grammar {
+			include "mod/security/Lenient-Encryption11-ghc.rnc" {
+				start = xenc_EncryptedData | xenc_EncryptedKey
+				xenc_EncryptionPropertyType =
+					mixed {
+						attribute Target { xsd:anyURI }?,
+						attribute Id { xenc_IdType }?,
+						attribute xml:* { text }*,
+						xenc_EncryptionPropertyContent
+					}
+				xenc_anyForeignElement =
+					element * - (xenc:* | epub_xenc_compress:*) {
+						mixed { security_anyAttribute*, security_anyElement* }
+					}
+			}
+			xenc_EncryptionMethodOtherParams |= attribute Algorithm { xsd:anyURI "http://www.idpf.org/2008/embedding" }
+			xenc_EncryptionAlgorithms |= xsd:anyURI "http://www.idpf.org/2008/embedding"
+			xenc_EncryptionPropertyContent = (epub_xenc_Compression & xenc_anyForeignElement*) | xenc_anyForeignElement+
+			epub_xenc_Compression = element epub_xenc_compress:Compression { epub_xenc_CompressionType }
+			epub_xenc_CompressionType =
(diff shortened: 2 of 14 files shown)


danielweck:
Unfortunately this RNC does not work in practice (let me know if I am wrong though, I am no RelaxNG expert).

```
default namespace = "urn:oasis:names:tc:opendocument:xmlns:container"
namespace epub_xenc_compress = "http://www.idpf.org/2016/encryption#compression"

start =
    element encryption {
        grammar {
            include "http://www.w3.org/TR/2013/NOTE-xmlsec-rngschema-20130411/Lenient-Encryption11-ghc.rnc" {
                start = xenc_EncryptedData | xenc_EncryptedKey
            }
            xenc_EncryptionMethodOtherParams |= attribute Algorithm { xsd:anyURI "http://www.idpf.org/2008/embedding" }
            xenc_EncryptionAlgorithms |= xsd:anyURI "http://www.idpf.org/2008/embedding"
            xenc_EncryptionProperty_child |= epub_xenc_Compression?
            epub_xenc_Compression = element epub_xenc_compress:Compression { epub_xenc_CompressionType }
            epub_xenc_CompressionType =
                attribute Method { "0" | "8" },
                attribute OriginalLength { xsd:nonNegativeInteger }
        }+
}
```

First of all, `xenc_EncryptionProperty_child` is not defined in the included `Encryption11-ghc.rnc`, this should in fact be `xenc_EncryptionPropertyType` which is defined as:

```
xenc_EncryptionPropertyType =
  mixed {
    attribute Target { xsd:anyURI }?,
    attribute Id { xenc_IdType }?,
    attribute xml:* { text }*,
    xenc_anyForeignElement+
  }
```

However, using the `|=` operator to refine these validation rules doesn't work either, because `xenc_anyForeignElement` "swallows" every children as valid. Furthermore, elements are ordered (unlike attributes), so there is this issue as well.

On a side note, this remote URL:
`include "http://www.w3.org/TR/2013/NOTE-xmlsec-rngschema-20130411/Lenient-Encryption11-ghc.rnc"`
...should be replaced with the local asset:
`include "mod/security/Lenient-Encryption11-ghc.rnc"`

So, I came up with this solution:

```
default namespace = "urn:oasis:names:tc:opendocument:xmlns:container"
namespace epub_xenc_compress = "http://www.idpf.org/2016/encryption#compression"
namespace xenc = "http://www.w3.org/2001/04/xmlenc#"

start =
    element encryption {
        grammar {
            include "mod/security/Lenient-Encryption11-ghc.rnc" {
                start = xenc_EncryptedData | xenc_EncryptedKey

                xenc_EncryptionPropertyType =
                  attribute Target { xsd:anyURI }?,
                  attribute Id { xenc_IdType }?,
                  attribute xml:* { text }*,
                  mixed {
                    (epub_xenc_Compression, xenc_anyForeignElement*) | xenc_anyForeignElement_+
                  }
            }
            xenc_EncryptionMethodOtherParams |= attribute Algorithm { xsd:anyURI "http://www.idpf.org/2008/embedding" }
            xenc_EncryptionAlgorithms |= xsd:anyURI "http://www.idpf.org/2008/embedding"

            xenc_anyForeignElement_ = element * - (xenc:* | epub_xenc_compress:Compression) { mixed { security_anyAttribute*, security_anyElement* } }

            epub_xenc_Compression = element epub_xenc_compress:Compression { epub_xenc_CompressionType }
            epub_xenc_CompressionType =
                attribute Method { "0" | "8" },
                attribute OriginalLength { xsd:nonNegativeInteger }
        }+
}
```

This allows the `Compression` element with optional `xenc_anyForeignElement` following siblings (not before, only after). This also continues to allow `xenc_anyForeignElement` under the condition that there are no `Compression` elements amongst children.

Example tests (trimmed XML for brevity):


```xml
      <EncryptionProperty>
        <DummyBefore xmlns="http://www.dummy.org/fake/namespace" attr="val"></DummyBefore>
        <Compression xmlns="http://www.idpf.org/2016/encryption#compression" Method="0" OriginalLength="A"></Compression>
      </EncryptionProperty>
```

Expected errors:
* 'Error while parsing file: element "Compression" not allowed here; expected the element end-tag or text' (that's because of `DummyBefore`)
* 'Error while parsing file: value of attribute "OriginalLength" is invalid; must be an integer' (it is `A`)

```xml
      <EncryptionProperty>
        <Compression xmlns="http://www.idpf.org/2016/encryption#compression" Method="9" OriginalLength="0"></Compression>
        <DummyAfter xmlns="http://www.dummy.org/fake/namespace" attr="val"></DummyAfter>
      </EncryptionProperty>
```

Expected errors:
* 'Error while parsing file: value of attribute "Method" is invalid; must be equal to "0" or "8"' (it is `9`)
* (note that `DummyAfter` is allowed here)

```xml
      <EncryptionProperty>
        <DummyBefore xmlns="http://www.dummy.org/fake/namespace" attr="val"></DummyBefore>
        <DummyAfter xmlns="http://www.dummy.org/fake/namespace" attr="val"></DummyAfter>
      </EncryptionProperty>
```

Expected errors:
* (none, this is perfectly valid)

murata2makoto:
I am responsible for the not-so-easy-to-customize design of xenc-schema.rnc.  I was involved in the design of RELAX NG, NVDL, and the RNG schemas for XML security.

@danielweck You appear to disallow elder brothers of the Compression element on purpose.  Is this backed by the current draft of OCF 3.2?

The http URLs do not cause network access because epubcheck use XML catalog (epub30-catalog.xml).  So, no changes for schema locations are needed.

Finally, I would like to slightly modify the schema for consistency with other security RNG schemas.  Here is my proposal.

```
default namespace = "urn:oasis:names:tc:opendocument:xmlns:container"
namespace xenc = "http://www.w3.org/2001/04/xmlenc#"
namespace epub_xenc_compress = "http://www.idpf.org/2016/encryption#compression"

start =
    element encryption {
        grammar {
            include "http://www.w3.org/TR/2013/NOTE-xmlsec-rngschema-20130411/Lenient-Encryption11-ghc.rnc" {
                start = xenc_EncryptedData | xenc_EncryptedKey
                xenc_EncryptionPropertyType =
                  mixed {
                    attribute Target { xsd:anyURI }?,
                    attribute Id { xenc_IdType }?,
                    attribute xml:* { text }*,
                    xenc_EncryptionPropertyContent
                  }
                xenc_anyForeignElement = element * - (xenc:* | epub_xenc_compress:*) {
                  mixed { security_anyAttribute*, security_anyElement* } }
            }
            xenc_EncryptionMethodOtherParams |= attribute Algorithm { xsd:anyURI "http://www.idpf.org/2008/embedding" }
            xenc_EncryptionAlgorithms |= xsd:anyURI "http://www.idpf.org/2008/embedding"

            xenc_EncryptionPropertyContent = epub_xenc_Compression?, xenc_anyForeignElement+
	    
            epub_xenc_Compression = element epub_xenc_compress:Compression { epub_xenc_CompressionType }
            epub_xenc_CompressionType =
                attribute Method { "0" | "8" },
                attribute OriginalLength { xsd:nonNegativeInteger }
        }+
    }

```

mattgarrish:
I've updated the pull request per the discussions, but made one change to your suggested schema @murata2makoto. This line:

> xenc_EncryptionPropertyContent = epub_xenc_Compression?, xenc_anyForeignElement+

requires a foreign element, but my understanding of the compression element is that it should be allowed alone. I also can't find anything in the specification that requires the Compression element be in any specific position, so I replaced it this with:

> xenc_EncryptionPropertyContent = (epub_xenc_Compression & xenc_anyForeignElement*) | xenc_anyForeignElement+

But if either you or @danielweck sees anything wrong with this approach, let me know.

rdeltour:
@murata2makoto 
> The http URLs do not cause network access because epubcheck use XML catalog (epub30-catalog.xml). So, no changes for schema locations are needed.

In fact, I don't think EPUBCheck does use a catalog resolver! As far as I know this catalog file is mostly here to be used by third-party tools, e.g. to configure XML editors.

@danielweck the changes made by @mattgarrish look OK to me. Before I mere the PR, Can you confirm the bit about the `Compression` element position?

The only slight change I will make when integrating the PR is running the error assertions before the ones about the error content, to avoid an NPE if only 1 error is found. But that's a detail  :-)

src/test/java/com/adobe/epubcheck/ocf/OCFCheckerTest.java
```diff
@@ -384,4 +398,27 @@ public void testInvalidLoremSCH30()
...
+
+    if (2 != testReport.getErrorCount() || 0 != testReport.getWarningCount())
+    {
+      outWriter.println(testReport);
+    }
+    if (testReport.errorList.size() > 0)
+    {
+      assertTrue(testReport.errorList.get(0).message
+          .contains("value of attribute \"Method\" is invalid; must be equal to \"0\" or \"8\""));
+      assertTrue(testReport.errorList.get(1).message
```
rdeltour: this can possibly throw an NPE since we didn't assert at this point that there are at least two errors…

murata2makoto: >In fact, I don't think EPUBCheck does use a catalog resolver! As far as I know this catalog file is mostly here to be used by third-party tools, e.g. to configure XML editors.

Then, does epubcheck always retreive schemas?

rdeltour:
> does epubcheck always retreive schemas?

I may be wrong, but I think it uses local schemas and relative references everywhere.

danielweck:
This looks good to merge.

I opened a WIP PR to possibly improve the schema, and most probably to add tests: https://github.com/w3c/epubcheck/pull/978

Issue #896, opened 2018-11-12
aria-describedat does not exist


mattgarrish:
aria-describedat was dropped from the ARIA specification and is no longer in 3.2.

It needs to be removed from the schemas so it generates an error, as it is currently only being reported as deprecated.

rdeltour:
Done in #934, will be released in v4.2.0.

Issue #893, opened 2018-11-12
Allow the latest recommended SVG for content documents


mattgarrish:
The Content Documents specification uses an auto-updating reference to the latest W3C version of SVG so that changes can be immediately reflected in validation. Currently this is SVG 2.

Note that part of this change is that the previous restrictions on the use of animation elements and events is removed.

Reference: https://w3c.github.io/publ-epub-revision/epub32/spec/epub-contentdocs.html#sec-overview-relations-svg

rdeltour:
The schema update relies on the SVG schemas included in the Nu Html Checker, which is not strictly checking SVG 2, but adopts a more pragmatic approach ("There are otherwise no plans to add blanket support for SVG2 as a whole", see validator/validator#564).

I think we should follow the Nu Html Checker here (to the extent EPUBCheck can, i.e. currently limited to the schemas). This is what is done in #934.

mattgarrish:
And as noted in https://github.com/w3c/publ-epub-revision/issues/1219 the reference is technically still 1.1. I wasn't looking closely at the status of the linked SVG document, which is still only a CR.

But in any case, I agree we should follow the model adapted by the nu checker.

rdeltour:
OK, closing the issue for now. We can reopen ad hoc issues if user complain that valid SVG content is rejected by EPUBCheck.

Issue #892, opened 2018-11-12
Allow the latest recommended HTML for content documents


mattgarrish:
The Content Documents specification now uses an auto-updating reference to the latest W3C version of HTML so that changes can be immediately reflected in validation. Currently this is HTML 5.2.

Reference: https://w3c.github.io/publ-epub-revision/epub32/spec/epub-contentdocs.html#sec-overview-relations-html

rdeltour:
Done in #934, will be released in v4.2.0.

Issue #835, opened 2018-03-08
Invalid MathML markup not throwing an error


pkra:
Nested `math` tags should throw an error as they are not valid.

pkra:
> I tried to reproduce it several times and every time there was an error, hence all works as expected.

Argh, I can't reproduce this in an isolated minimal example and of course I fixed the sample where I ran into this and don't recall which of the expressions caused the issue.

Sigh. 

Sorry for the noise. I'll re-open if I run into this again.

pkra:
Got it. The problem is inside epub-switch elements. 

Something like 

```xml
<?xml version="1.0" encoding="utf-8"?>
<!DOCTYPE html>

<html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops">
<head>
  <title>nested math tags should be invalid</title>
</head>

<body>
 <h1>nested math tags should be invalid even inside epub switch</h1>
 <epub:switch>
   <epub:case required-namespace="http://www.w3.org/1998/Math/MathML">
    <math alttext="placeholder" display="block" overflow="scroll" xmlns="http://www.w3.org/1998/Math/MathML">
      <math alttext="a"><mn>11</mn>
       <mo>,</mo>
       <mo>−</mo>
       <mn>3</mn>
       <mo>,</mo>
       <mn>5</mn>
      </math>
    </math>
   </epub:case>
  <epub:default>
 <img src="../Images/Image0001.svg" alt=""/></epub:default>
</epub:switch>
</body>
</html>
```

With 4.0.2, this gets validated without errors.

Sample [nested-math-tags.zip](https://github.com/IDPF/epubcheck/files/2149169/nested-math-tags.zip)  (change extension to epub).

pkra:
I didn't test master so this may be fixed already.

rkwright:
Tested against epubcheck 4.0.3 built from master.  No errors.

rdeltour:
Done in #934

PR #934, opened 2019-01-09
Update schemas from upstream Nu Html Checker sources

rdeltour:
This is a major schema update, to reuse the latest schemas from
the Nu Html Checker, with the following tweaks:

- adapt the integration in schema drivers
- replace the Nu Html Checker's pluggable datatype library with
  EPUBCheck’s more limited static RelaxNG datatypes
- add EPUB-specific content model extensions

Changes specific to XHTML (#892):

- keep allowing missing `title` in the `head` element (#448)
- but report missing `title` as a WARNING (with a Schematron check)
- keep allowing `meta http-equiv` in encoding declaration state
  (allowed in W3C but forbidden in WHATWG's HTML standard)
- keep allowing the `border` attribute on `table` elements
  (allowed in W3C but forbidden in WHATWG's HTML standard)
- report `aria-describedat` as a schema error (#896)
- add tests for several obsolete features which appeared in previous
  test content (`time/@pubdate`, `iframe/@seamless`, `keygen`, `menu`,
  `dropzone`, `contextmenu`, in-body `style`, `style/@Scoped`)
-  check MathML and XHTML content in `epub:switch` (#835)

Changes specific to MathML:

- use the schemas from Nu Html Checker, which are based on older
  MathML schema but bring some specific tweaks.
- re-add the changes from the newer MathML 3.0 (2nd edition) schemas
- improve the schema checks of EPUB-specific `annotation-xml`
  restrictions (notably check the combination of the `name` and
  `encoding` attributes)

Changes specific to SVG (#779, #893):

- use the schemas from Nu Html Checker, which are based SVG 1.1

Changes specific to Navigation Documents:

- do not check the presence of mandatory headings on `nav` elements
  with no `epub:type` attribute


danielweck:
Related PR:
https://github.com/w3c/epubcheck/pull/927

murata2makoto:
Why do we need inkscape-draft.rnc?

rdeltour:
> Why do we need inkscape-draft.rnc?

TBH I'm not sure. For now I "just" copied over the schemas used and referenced by the Nu Html Checker. In fact I was wondering the same, but haven't had the time to look it up in detail yet.

I'm assuming it contains pattern definitions useful to general SVG checking, but again to be taken with a grain of salt before a more detailed look.

murata2makoto:
Do you plan to introduce custom Java codes in  [Nu Html Checker ](https://github.com/validator/validator) as well?  In my understanding, such codes and schemas in the Nu Html Checker are intended to be used together.  However, even if epubcheck borrows schemas only, I suppose that the result will be already a big improvement.

rdeltour:
> Do you plan to introduce custom Java codes in [Nu Html Checker ](https://github.com/validator/validator) as well?

Unfortunately, a full integration of the Nu Html Checker is out of scope for v4.2.0, which aims at bringing EPUB 3.2 support to the community as soon as possible.

That said, I totally agree that there's only so much the schemas can do, and a (very) significant part of the Nu Html Checker’s functionality resides in its Java code. We do intend to experiment with a full integration, later down the road, but it’s a more ambitious refactoring.

For now, as you say the schemas are already a big improvement!

mattgarrish:
One more spec inconsistency I found validating the samples: W3C allows the `border` attribute on tables but the schemas/WHATWG do not.

I think we need to override and allow, as I suspect this could affect a lot of people.

Otherwise, the schemas caught all the same issues as before plus a lot of new legitimate problems. Definitely looking like a good upgrade.

rdeltour:
> One more spec inconsistency I found validating the samples: W3C allows the `border` attribute on tables but the schemas/WHATWG do not.
> 
> I think we need to override and allow, as I suspect this could affect a lot of people.

Ok, I'll override this.

I'm wondering how we could get reasonably representative stats on existing EPUB content. That would be super useful.

PR #944, opened 2019-01-17
fix: `switch` item property doesn't apply to `svg:switch`

### LINKED ISSUE #857: declaring svg:switch in package file?
While investigating the use of SVG's `requiredExtensions` attribute in EPUB 3.2, I tried some experiments with SVG `switch` and `foreignResource`. Here's a content document:

```html
<?xml version="1.0" encoding="UTF-8"?>
<html xmlns="http://www.w3.org/1999/xhtml"
      xmlns:epub="http://www.idpf.org/2007/ops"
      xml:lang="en-US" lang="en-US"
      xmlns:svg="http://www.w3.org/2000/svg">
   <head>
      <title>SVG SWITCH</title>
      <meta charset="utf-8" />
   </head>
   <body>
      <svg:svg width="400px" height="300px" viewBox="0 0 400 300">
         <svg:switch>
            <svg:foreignObject width="100" height="50" requiredExtensions="http://www.idpf.org/2007/ops">
               <p xmlns="http://www.w3.org/1999/xhtml">foo</p>
            </svg:foreignObject>
            <svg:text font-size="10" font-family="Verdana">
               <svg:tspan x="10" y="10">bar</svg:tspan>
            </svg:text>
         </svg:switch>
      </svg:svg>
      <section id="d10e42">
         <h1 id="d10e44">SVG SWITCH</h1>
         <p>Hello <a href="#d10e44">there</a></p>
      </section>
   </body>
</html>
```

1. EPUBCheck wants me to declare the `switch` element in the package file even though this is `svg:switch` and not `epub:switch`:

```
ERROR(OPF-014): ..epub/EPUB/georgia.xhtml(-1,-1): The property 'switch' should be declared in the OPF file.
```

2. Any ordinary xhtml link that occurs **after** the SVG brings up an error about needing a title:

```
WARNING(ACC-011): ..epub/EPUB/georgia.xhtml(23,46): SVG hyperlinks should have a human-readable title (using the 'title' child element or the 'xlink:title' attribute).
```

Of course, these are not SVG hyperlinks.

rdeltour:
The `switch` property of Package Documents `item` elements indicates
that the described Publication Resource contains one or more instances
of the `epub:switch` element. It should not be declared when the EPUB
contains SVG using the `svg:switch` element.

Fixes #857

### DIFF
--- a/src/main/java/com/adobe/epubcheck/ops/OPSHandler30.java
+++ b/src/main/java/com/adobe/epubcheck/ops/OPSHandler30.java
@@ -245,7 +245,7 @@ public class OPSHandler30 extends OPSHandler
     {
       requiredProperties.add(ITEM_PROPERTIES.SCRIPTED);
     }
-    else if (!context.mimeType.equals("image/svg+xml") && name.equals("switch"))
+    else if (EpubConstants.EpubTypeNamespaceUri.equals(e.getNamespace()) && name.equals("switch"))
     {
       requiredProperties.add(ITEM_PROPERTIES.SWITCH);
     }
--- a/src/test/java/com/adobe/epubcheck/api/Epub30CheckExpandedTest.java
+++ b/src/test/java/com/adobe/epubcheck/api/Epub30CheckExpandedTest.java
@@ -419,6 +419,13 @@ public class Epub30CheckExpandedTest extends AbstractEpubCheckTest
   }
 
   @Test
+  public void testValidateEPUB30_svgSwitch()
+  {
+    // tests that svg:switch doesn't trigger the OPF 'switch' property check
+    testValidateDocument("valid/svg-switch/");
+  }
+
+  @Test
   public void testValidateEPUB30_videoAudioTrigger()
   {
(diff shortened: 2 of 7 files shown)


src/main/java/com/adobe/epubcheck/ops/OPSHandler30.java
```diff
@@ -245,7 +245,7 @@ else if (name.equals("script"))
     {
       requiredProperties.add(ITEM_PROPERTIES.SCRIPTED);
     }
-    else if (!context.mimeType.equals("image/svg+xml") && name.equals("switch"))
+    else if (EpubConstants.HtmlNamespaceUri.equals(e.getNamespace()) && name.equals("switch"))
```
mattgarrish: Won't this look for a switch element in the XHTML namespace?

I think you want EpubTypeNamespaceUri.

rdeltour: Doh. Of course 😅.

PR #946, opened 2019-01-18
fix: valid JPEG files were considered 'corrupted'

### LINKED ISSUE #850: Corrupted image file encountered.
I have an epub which gives no error when validted by epubcheck.
But I wanted to decrease the epub size, so I compressed the jpeg images in Calibre with lossy compression. After that  epubcheck gives an error message for every jpeg file: "Corrupted image file encountered."

See the attached image as an example for "corrupted image file".

w=mc^2

rdeltour:
Improves the JPEG parsing by updating the TwelveMonkeys ImageIO plugin
to latest version 3.4.1.

Fixes #850


rdeltour:
Merged in c4a2cfac

PR #947, opened 2019-01-18
Re-enable openjdk11 on Travis CI

rdeltour:
The builds are now run on `xenial`, where openjdk 10 and 11 are pre-installed.

See also this [Travis CI community discussion](https://travis-ci.community/t/install-of-openjdk11-is-failing/1914).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: java
 jdk:
+  - openjdk11
   - oraclejdk11
   - openjdk10
   - oraclejdk9


rdeltour:
Apparently `xenial` doesn't support installing JDKs below Java 9, and even then it fails to download `oraclejdk9`.

For reference, Travis CI uses the [`install-jdk.sh`](https://github.com/sormuras/bach/blob/master/install-jdk.sh) script to fetch non-default JDKs.

Issue #875, opened 2018-11-12
Remove epubsc from reserved prefixes list


mattgarrish:
Scriptable components have been shelved, so the reserved prefix should be removed from the reserved list. If the prefix is used, it has to be declared.

Reference: https://w3c.github.io/publ-epub-revision/epub32/spec/epub-packages.html#sec-metadata-reserved-prefixes

rdeltour:
Done in #952.

PR #951, opened 2019-01-21
remove epubsc from reserved prefixes

mattgarrish:
This removes epubsc from the reserved prefixes, as noted in #875, and adds it to the known vocabularies so that property names can still be verified.

I could be easily swayed that we should just remove all traces of scriptable components entirely, though, as it seems very remote that those particular specifications will ever be finished in the post-IDPF world.


mattgarrish:
Let me try this one again...

PR #952, opened 2019-01-21
Remove epubsc from reserved prefixes

mattgarrish:
This removes epubsc from the reserved prefixes, per issue #875, and adds it to the known vocabularies so that property names can still be verified.

I could be easily swayed that we should just remove all traces of scriptable components entirely, though, as it seems very remote that those particular specifications will ever be finished in the post-IDPF world.


mattgarrish:
Sure, if you wan't to ensure no one accidentally turns it back on I can add one.

rdeltour:
rebased and merged as 02397ec5

Issue #874, opened 2018-11-12
Add support for application/javascript


mattgarrish:
application/javascript replaces text/javascript as the preferred media type for scripts in the manifest.

Reference: https://w3c.github.io/publ-epub-revision/epub32/spec/epub-spec.html#cmt-grp-content-documents

rdeltour:
Done in #954

PR #954, opened 2019-01-22
Add application/javascript to CMTs

mattgarrish:
Add "application/javascript" for identifying javascript in the package doc, per issue #874 

Also consolidates checking of script media types.


rdeltour:
Merged as f4566b63

Issue #775, opened 2017-07-02
Yearless dates are valid HTML but flagged


robinwhittleton:
[HTML5.1 allows yearless dates](https://www.w3.org/TR/html51/infrastructure.html#yearless-dates), but epubcheck currently flags them. It looks like amending the current set of regexes in datatypes.rnc to include something like `([0-9]{2})-([0-9]{2})` would fix it but this is the first time I’ve looked at the code so I’ll defer unless this doesn’t get picked up.

(somewhat similar to https://github.com/IDPF/epubcheck/issues/341)

tofi86:
AFAIK EPUB 3 specifies a specific HTML 5 subset which is not HTML 5.1 (am I right @mattgarrish?) and the newer HTML versions are only supported by EPUB 3.1 which however isn't supported by EpubCheck yet.

robinwhittleton:
Thanks for the reply. HTML5.0 also defines [a yearless date format](https://www.w3.org/TR/html5/infrastructure.html#valid-yearless-date-string) as [valid for the datetime attribute](https://www.w3.org/TR/html5/text-level-semantics.html#datetime-value), and it looks like EPUB 3.0 pretty much [wholesale imports the W3C HTML5.0 spec](http://www.idpf.org/epub/301/spec/epub-overview.html#refContentDocs3) (assuming XHTML serialisation)? Sorry if I’m getting this wrong, I’m pretty new to EPUB stuff.

mattgarrish:
Looks like a bug. From what I see, there's no accounting for yearless dates in the date.or.time datatype. I wouldn't modify html.date as that looks like it would allow in input/date which requires a year from my reading. Probably need a new date type like:

datatype.html5.date.or.time = datatype.html5.date | datatype.html5.month | datatype.html5.yearless | datatype.html5.time | datatype.html5.datetime.global

datatype.html5.yearless = xsd:token { pattern ='(--)?(0[1-9]|1[0-2])-(0[0-9]|[1-2][0-9]|3[0-1])' }

(All the other date patterns use simple [0-9] ranges, so maybe the above is a bit of overkill since it will still allow bad dates.)

rdeltour:
Closed in #957

PR #957, opened 2019-01-22
improve handling of dates/times/durations

mattgarrish:
Per the report in #775, the handling of dates and times is much more limited in epubcheck than html allows. This PR isn't going to fix all the issues, as some checks can really only be programmed; it's more an interim step to future integration.

What it does is make the following changes to better align with the html definitions:

- adds support for yearless dates, year-only dates, and durations for `time/datetime`
- allows a space in addition to 'T' for separating dates and times
- makes the colon optional in timezone offsets
- limits seconds to three digits after a period

I checked the valid/invalid files against validator.nu and the only discrepancy I found is that it doesn't recognize a yearless date that begins with two hyphens. It looks like a bug on that end, though, as the HTML specification clearly says [two optional hyphens are allowed at the start](https://www.w3.org/TR/html/infrastructure.html#yearless-dates).


rdeltour:
rebased and merged as ef5b94c9

PR #966, opened 2019-02-07
Set reported EPUB 3.x version to "3.2"

### LINKED ISSUE #943: Report 3.x validation as following the EPUB 3.2 rules
It current says:

```
Validating using EPUB version 3.0.1 rules
```

Thanks for the catch @murata2makoto!

rdeltour:
Fix #943


rdeltour:
> except having to change every test every time seems like a nuisance.

damn right 😭 (if only we weren't on a tight schedule and were able to massively cleanup the test suite before implementing all this! 🙃)

PR #968, opened 2019-02-08
normalize HTML case-insensitive attributes + basic support for custom elements

### LINKED ISSUE #932: Add basic support for HTML custom elements
Hi, 

Out of curiosity, I’ve been playing with custom elements v1 a little bit (so this is [the HTML spec](https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements)). [Support isn’t necessarily bad in real life](https://caniuse.com/#feat=custom-elementsv1) and well, although I’m not advising anyone to use that in production, I could probably see workflow/internal use cases – especially as an app that is using a WebView that supports that, supports that.

So I’m wondering… is this something epubcheck has to align with? I am currently getting the following error: 

```
ERROR | RSC-005 | /OEBPS/Text/002_otf.xhtml (line 19, col 12) | Error while parsing file 'element "dink-us" not allowed here; expected the element end-tag, text or element "a", "abbr", "address", "area", "article", "aside", "audio", "b", "bdi", "bdo", "blockquote", "br", "button", "canvas", "cite", "code", "command", "datalist", "del", "details", "dfn", "div", "dl", "em", "embed", "epub:switch", "epub:trigger", "fieldset", "figure", "footer", "form", "h1", "h2", "h3", "h4", "h5", "h6", "header", "hgroup", "hr", "i", "iframe", "img", "input", "ins", "kbd", "keygen", "label", "link", "main", "map", "mark", "menu", "meta", "meter", "nav", "ns1:math", "ns2:svg", "object", "ol", "output", "p", "pre", "progress", "q", "ruby", "s", "samp", "script", "section", "select", "small", "span", "strong", "style", "sub", "sup", "table", "textarea", "time", "u", "ul", "var", "video" or "wbr" (with xmlns:ns1="http://www.w3.org/1998/Math/MathML" xmlns:ns2="http://www.w3.org/2000/svg")'.
```

It is my understanding this could be more of a parsing error, probably because of the dash, but I could be wrong. 

On a related note, you can add contents and attributes with `connectedCallback()` (cf. [this demo](https://codepen.io/JayPanoz/pen/zyZEMM) for instance) so I’m not sure how you can handle that if you don’t check for markup added with JS already – sorry, didn’t bother to try that.

### LINKED ISSUE #941: Normalize the case of boolean attributes values
epubcheck flags an error for the following attributes on element iframe: 

- allowFullScreen="allowFullScreen"
- allowfullscreen=""
- allow="fullscreen"

Error reads as follows: 

```
ERROR(RSC-005): ./Documents/Work/Validator/Testing Blank Files/S_EPUB_AllowFullScreen_ALL.epub/OEBPS/xhtml/jac_9781319048860_back.xhtml(18,106): Error while parsing file: attribute "allowFullScreen" not allowed here; expected attribute...
```

("..." omits the enumerated allowed attributes on iframe).

Markup from file here for testing: 

```
<iframe src="http://www.youtube.com/embed/xDMP3i36naA" allowFullScreen="allowFullScreen"></iframe>
<iframe src="http://www.youtube.com/embed/xDMP3i36naA" allowfullscreen=""></iframe>
<iframe src="http://www.youtube.com/embed/xDMP3i36naA" allow="fullscreen"></iframe>
```

HTML validator allows these, so I propose that this may be a bug in epubcheck.

rdeltour:
## parse boolean/enumerated HTML attributes as case-insensitive
- add an `HTMUtils` facility to get whether an HTML attribute has a
  case-insensitive value (boolean and enumerated attributes)
- pre-process case-insensitive HTML attributes to lower-case their value
- refactor the pre-processing attribute logic for more clarity
- add a test

Fix #941

## basic schema support for HTML custom elements

- add an `isCustomElement` method in `HTMLUtils`.
- pre-process custom elements in `XMLParser` to put them in the
  proprietary namespace used by the Nu Html Checker schemas to treat
  them distinctively.
- add a simple test case.

Fix #932

### DIFF
--- a/src/main/java/com/adobe/epubcheck/xml/HTMLUtils.java
+++ b/src/main/java/com/adobe/epubcheck/xml/HTMLUtils.java
@@ -75,4 +75,9 @@ public class HTMLUtils
 
   }
 
+  public static boolean isCustomElement(String namespace, String name)
+  {
+    return Namespaces.XHTML.equals(namespace) && Preconditions.checkNotNull(name).contains("-");
+  }
+
 }
--- a/src/main/java/com/adobe/epubcheck/xml/HTMLUtils.java
+++ b/src/main/java/com/adobe/epubcheck/xml/HTMLUtils.java
@@ -1,0 +1,78 @@
+package com.adobe.epubcheck.xml;
+
+import com.google.common.base.Preconditions;
+
+/**
+ * Utilities for HTML-specific logic.
+ * 
+ */
+public class HTMLUtils
+{
+  /**
+   * Returns whether an attribute is defined as having a case-insensitive value in
+   * HTML. This is notably the case of boolean attributes and enumerated
+   * attributes.
+   * 
+   * @param name
+   *          the name of an attribute defined in the HTML specification
+   * @return <code>true</code> iff the attribute value is case-insensitive
+   */
+  public static boolean isCaseInsensitiveAttribute(String name)
+  {
+    switch (Preconditions.checkNotNull(name))
+    {
+    case "align":
+    case "allowfullscreen":
+    case "allowpaymentrequest":
+    case "allowusermedia":
+    case "async":
+    case "autocapitalize":
+    case "autocomplete":
+    case "autofocus":
+    case "autoplay":
+    case "checked":
+    case "contenteditable":
+    case "controls":
+    case "crossorigin":
+    case "default":
+    case "defer":
+    case "dir":
+    case "disabled":
+    case "draggable":
+    case "formnovalidate":
+    case "hidden":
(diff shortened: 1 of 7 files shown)


rdeltour:
> The only thing I noticed is that custom element names aren't supposed to allow uppercase ascii characters, but this isn't tested. The same is true with the online validators, though.

Right: I applied the same (naïve) filtering as done in validator.nu. I don't know if they do some more advanced element name validation via Java assertions however.

In any case, this already gives us basic support, and I think we can safely wait for a higher integration with validator.nu to do something more advanced. As Jiminy said, I don't expect people to use custom elements in production anytime soon ;-)

Issue #859, opened 2018-10-05
XSLT warning "The parent axis starting at a document node will never select anything"


danielweck:
Hello!

* I am on MacOS (latest)
* I installed the Oracle JDK (latest v11.0.1 from https://www.oracle.com/technetwork/java/javase/downloads/jdk11-downloads-5066655.html )
* I installed Maven (latest)

* I cloned this repository and checkout'ed the `master` branch
* I invoked the Maven install command (no problems)
* I ran EpubCheck from the command line against a few EPUBs (seems to work just fine).
* However, the following message is reported over 20 times in the console (just before the actual validation message):

```
Warning at char 3 in xsl:param/@select on line 459 column 43 
  SXWN9000: The parent axis starting at a document node will never select anything
```

```
Validating using EPUB version 3.0.1 rules.
No errors or warnings detected.
Messages: 0 fatal / 0 errors / 0 warnings / 0 info
```

System:

`java --version`
```
java 11 2018-09-25
Java(TM) SE Runtime Environment 18.9 (build 11+28)
Java HotSpot(TM) 64-Bit Server VM 18.9 (build 11+28, mixed mode)
```

`mvn --version`
```
Apache Maven 3.5.4 (1edded0938998edf8bf061f1ceb3cfdeccf443fe; 2018-06-17T19:33:14+01:00)
Maven home: /Applications/apache-maven
Java version: 11, vendor: Oracle Corporation, runtime: /Library/Java/JavaVirtualMachines/jdk-11.jdk/Contents/Home
Default locale: en_GB, platform encoding: UTF-8
OS name: "mac os x", version: "10.14", arch: "x86_64", family: "mac"
```

`java -jar target/epubcheck.jar --version`
```
EpubCheck v4.0.3-SNAPSHOT
No file specified in the arguments. Exiting.
epubcheck completed
```

rdeltour:
> Warning at char 3 in xsl:param/@select on line 459 column 43 SXWN9000: The parent axis starting at a document node will never select anything

Would there be a file associated somewhere in the log with this line and column number?
If not we'll have to do some spelunking :-)

tofi86:
Note that there are no Travis checks for Java 11 yet. We should set those up and check if it’s a Java 11 issue.

danielweck:
[...]

Issue #491, opened 2014-12-28
SVG with custom namespace fails check


acabal:
Checking a SVG using a custom namespace fails with an error--epubcheck doesn't seem to recognize the custom namespace.  Here's a minimal, valid test SVG:

```
<?xml version="1.0" encoding="UTF-8"?>
<svg xmlns="http://www.w3.org/2000/svg" xmlns:test="https://example.org" version="1.1" viewBox="0 0 10 10" test:foo="bar">
</svg>
```

Expected result:
The valid SVG should not generate check errors.

Actual result:
Error while parsing file 'attribute "test:foo" not allowed here; expected attribute "alignment-baseline", "baseProfile", "baseline-shift", "class", "clip", "clip-path", "clip-rule", "color", "color-interpolation", "color-interpolation-filters", "color-profile", "color-rendering", "contentScriptType", "contentStyleType", "cursor", "direction", "display", "dominant-baseline", "enable-background", "externalResourcesRequired", "fill", "fill-opacity", "fill-rule", "filter", "flood-color", "flood-opacity", "font-family", "font-size", "font-size-adjust", "font-stretch", "font-style", "font-variant", "font-weight", "glyph-orientation-horizontal", "glyph-orientation-vertical", "height", "id", "image-rendering", "kerning", "letter-spacing", "lighting-color", "marker-end", "marker-mid", "marker-start", "mask", "ns:type", "onabort", "onactivate", "onclick", "onerror", "onfocusin", "onfocusout", "onload", "onmousedown", "onmousemove", "onmouseout", "onmouseover", "onmouseup", "onresize", "onscroll", "onunload", "onzoom", "opacity", "overflow", "pointer-events", "preserveAspectRatio", "requiredExtensions", "requiredFeatures", "shape-rendering", "stop-color", "stop-opacity", "stroke", "stroke-dasharray", "stroke-dashoffset", "stroke-linecap", "stroke-linejoin", "stroke-miterlimit", "stroke-opacity", "stroke-width", "style", "systemLanguage", "text-anchor", "text-decoration", "text-rendering", "unicode-bidi", "visibility", "width", "word-spacing", "writing-mode", "x", "xml:base", "xml:lang", "xml:space", "y" or "zoomAndPan" (with xmlns:ns="http://www.idpf.org/2007/ops")'.

chrisauclair:
[...]

PR #970, opened 2019-02-08
Allow foreign ns elements and attributes in svg content docs

mattgarrish:
(Resubmitting branched off next/v4.2.0 this time.) 

I believe this NVDL accomplishes what is needed for #491. It attaches XHTML elements as well as epub:, xlink: and xml: attributes, as well as attributes in no namespace. Elements and attributes from any other namespace are blanket allowed.

The empty pattern in the sch file is annoying, but without it oxygen was complaining that a required pattern was missing. Seems to be harmless, but feel free to comment it out.


murata2makoto:
I reviewed the NVDL script.  It looks good to me, but I would like to use a mode name different from "attach".  This mode does not attach foreign elements or attributes.

mattgarrish:
What about calling it allowForeignNS?

rdeltour:
I updated the PR:

- added code in the XML Parser to allow the `data-*` attributes by removing them at parsing time
- tweaked the NVDL to attach all namespaces to the Schematron validation
- tweaked the NVDL to set a stricter HTML-only mode to the RelaxNG validation in the context of a `foreignObject` element.

I took the opportunity to update Jing to v20181222 (which fixes #859 without us having to provide a hack, and even from an NVDL schema factory).

@mattgarrish I couldn't assign you to review your own PR, but if you can have a quick look… 😉

mattgarrish:
LGTM!

Issue #899, opened 2018-11-12
Remove warnings when position fixed or absolute are used in CSS


mattgarrish:
The recommendation not to use position: [fixed|absolute] was dropped from the CSS profile in 3.2 so detection of these values should not result in a warning anymore. (It was replaced with a note.)

Reference: https://w3c.github.io/publ-epub-revision/epub32/spec/epub-contentdocs.html#sec-css-content-conf

clapierre:
This is currently causing us some issues where we want to have a Screen Reader specific spoken text hidden off-screen with a different visual presentation of the text onscreen.
IE. a fancy chapter H1 of "One" with sub chapter heading of "Back to Basics"
but we want the screen reacher to hear
"Chapter One, Back to Basics"
IE: CSS 
 .sr-only
            {
              clip: rect(1px, 1px, 1px, 1px);
              **position: absolute !important;** 
              white-space: nowrap;
              height: 1px;
              width: 1px;
              overflow: hidden;
            }

So having this ability to add this CSS without EPUBCheck issuing a warning would be great.

rdeltour:
Note: a side effect of removing these warnings will be to fix the issue that they were also reported for EPUB 3.2 content, which didn't have this recommendation in the first place (only a "strongly discouraged" note).

rdeltour:
Fixed in #971.

PR #971, opened 2019-02-11
Change css position warnings to usage messages

mattgarrish:
This PR addresses #899 in the following ways:

- lowers CSS-006 and CSS-017 from WARNING to USAGE
- deletes CSS-027 as it becomes redundant with -017

Making these changes resulted in no warnings in the command line tests, though, so to add one back I deleted the title from one of the XHTML content documents. We don't have any CSS warnings in the specification, and my limited imagination couldn't come up with any other warnings for HTML.

### DIFF
--- a/src/main/resources/com/adobe/epubcheck/messages/MessageBundle.properties
+++ b/src/main/resources/com/adobe/epubcheck/messages/MessageBundle.properties
@@ -40,7 +40,7 @@ CSS_002=Empty or NULL reference found.
 CSS_003=Only UTF-8 and UTF-16 encodings are allowed, detected %1$s.
 CSS_004=Only UTF-8 and UTF-16 encodings are allowed, detected %1$s BOM.
 CSS_005=Conflicting alternate style attributes found: %1$s.
-CSS_006=CSS position:fixed property should not be used in EPUBs.
+CSS_006=CSS selector specifies fixed position.
 CSS_007=Font-face reference %1$s refers to non-standard font type %2$s.
 CSS_008=An error occurred while parsing the CSS: %1$s.
 CSS_009=Use of certain CSS such as Columns, Transforms, Transitions, box-sizing or KeyFrames can cause pagination issues.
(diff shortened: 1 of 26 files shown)


src/main/resources/com/adobe/epubcheck/messages/MessageBundle.properties
```diff
@@ -63,7 +63,6 @@ CSS_024=CSS class Selector is not used.
 CSS_024_SUG=Remove unused CSS selectors.
 CSS_025=CSS class Selector could not be found.
 CSS_025_SUG=Check for typos or define a class selector to document the use of the class.
-CSS_027=CSS selector specifies absolute position.
```
rdeltour: I'm wondering if we should "shift" the other messages, i.e. s/CSS_028/CSS_027.
Removing messages is quite rare, I don't think this happened before.

These message codes always felt a bit odd –at best arbitrary–; maybe we can leave it like that for now, and revisit this when we refactor after v4.2.0.

src/test/java/com/adobe/epubcheck/test/command_line_Test.java
```diff
@@ -233,7 +233,7 @@ public void xmpfile_Test()
   public void failonwarnings_Test()
   {
     String[] extraArgs = {"--failonwarnings"};
-    runExtraCommandLineArgTest("failonwarnings", 1, extraArgs);
+    runExtraCommandLineArgTest("failonwarnings", 0, extraArgs);
```
rdeltour: Doesn't this change remove the usefulness of this test, which if I understand correctly checks that EPUBCheck fails when a `WARNING` is reported?

Should we use the "remove the title" trick here too?

src/main/resources/com/adobe/epubcheck/messages/MessageBundle.properties
mattgarrish: I wondered about that, but there are already a few gaps and I didn't want to introduce any unrelated problems. (-014, -018, and -026 are missing from the CSS messages, for example, and I see similar holes in HTM, OPF, etc.) I'd vote for leaving until we come up with a more logical system.

src/test/java/com/adobe/epubcheck/test/command_line_Test.java
mattgarrish: Oops, ya, I missed that it was also testing warning. Will fix.

PR #972, opened 2019-02-15
Allow font/* as blessed media types

### LINKED ISSUE #872: Core media type fonts
This is a placeholder issue until https://github.com/w3c/publ-epub-revision/issues/1158 is resolved. Either we need to update the allowed core media types for font/woff2 and application/font-sfnt, or we should stop enforcing fallbacks for font resources.

mattgarrish:
This PR fixes #872 by allowing resources declared as font/* to be recognized as core media types, since fonts don't require fallbacks.

### DIFF
--- a/src/main/java/com/adobe/epubcheck/opf/OPFChecker30.java
+++ b/src/main/java/com/adobe/epubcheck/opf/OPFChecker30.java
@@ -491,7 +491,13 @@ public class OPFChecker30 extends OPFChecker implements DocumentValidator
 
   public static boolean isBlessedFontType(String type)
   {
-    return type.equals("application/vnd.ms-opentype") || type.equals("application/font-woff")
+    return type.equals("font/otf")
+        || type.equals("font/ttf")
+        || type.equals("font/woff")
+        || type.equals("font/woff2")
+        || type.equals("application/font-sfnt")
+        || type.equals("application/font-woff")
+        || type.equals("application/vnd.ms-opentype")
         || type.equals("image/svg+xml");
   }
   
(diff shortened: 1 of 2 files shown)


src/main/java/com/adobe/epubcheck/opf/OPFChecker30.java
```diff
@@ -492,7 +492,7 @@ public static boolean isCommonVideoType(String type)
   public static boolean isBlessedFontType(String type)
   {
     return type.equals("application/vnd.ms-opentype") || type.equals("application/font-woff")
-        || type.equals("image/svg+xml");
+        || type.equals("image/svg+xml") || type.startsWith("font/");
```
rdeltour: Don't we also need to add `application/font-sfnt` to the list?

mattgarrish: Ah, good point. I've changed the woff test to startsWith "application/font-" just to be liberal (PFR fonts could always rise from the dead).

rdeltour:
So I checked the current implementation, and it actually seems that EPUBCheck is already permissive on font types, and already did not require them to be CMT.

The method `isBlessedFontType` is really only used to report `CSS-007` (`INFO`).

Given this, what do you think about actually testing for exact matches against the types listed as CMT, so that any other type (e.g. due to typo like `font/tttf`) is reported as an INFO?

rdeltour:
For reference, the test [`testFont_NonCoreMediaType`](https://github.com/w3c/epubcheck/blob/v4.2.0-alpha-1/src/test/java/com/adobe/epubcheck/api/Epub30CheckExpandedTest.java#L625) tests that [a non-CMT font](https://github.com/w3c/epubcheck/blob/v4.2.0-alpha-1/src/test/resources/30/expanded/valid/font-othermediatype/EPUB/package.opf#L13) is allowed.

mattgarrish:
All that for a usage message... 😭 

Ya, I suppose it makes sense just to report fonts that aren't core media types.

rdeltour:
> Ya, I suppose it makes sense just to report fonts that aren't core media types.

OK, I'll list them explicitly and merge the PR 😊

PR #973, opened 2019-02-18
Issue/902

GrayWolfMT:
Updated code to check for Widht/Height on fixed layout SVG elements.  If no width/height/viewBox are present, reworded version of HTM_048 presented.  If width/height present buy no viewBox, a warning is displayed.

Additional warning added for widht/height with % values as spec recommends using pixel values for these if viewBox not present.


src/main/java/com/adobe/epubcheck/ctc/EpubSVGCheck.java
```diff
@@ -123,9 +123,22 @@ void checkViewBox(String svgDocEntry, Document doc)
     {
       Element svgElement = (Element) n.item(i);
       String viewport = svgElement.getAttributeNS(svgNS, "viewBox");
-      if (viewport == null || viewport.length() == 0)
+      boolean viewportFound = (viewport != null && viewport.length() > 0);
+      String height = svgElement.getAttributeNS(svgNS,"height");
+      boolean heightFound = (height != null && height.length() > 0);
+      String width = svgElement.getAttributeNS(svgNS,"width");
+      boolean widthFound = (width != null && width.length() > 0);
+      //if (viewport == null || viewport.length() == 0)
```
rdeltour: this comment line should be safely removed :-)

src/main/java/com/adobe/epubcheck/ctc/EpubSVGCheck.java
```diff
@@ -123,9 +123,22 @@ void checkViewBox(String svgDocEntry, Document doc)
...
+      if (!viewportFound)
       {
-        report.message(MessageId.HTM_048, EPUBLocation.create(svgDocEntry, XmlDocParser.getElementLineNumber(svgElement), XmlDocParser.getElementColumnNumber(svgElement)));
+      	if (!heightFound || !widthFound){
+          report.message(MessageId.HTM_048, EPUBLocation.create(svgDocEntry, XmlDocParser.getElementLineNumber(svgElement), XmlDocParser.getElementColumnNumber(svgElement)));
+		}else{
+          report.message(MessageId.HTM_054, EPUBLocation.create(svgDocEntry, XmlDocParser.getElementLineNumber(svgElement), XmlDocParser.getElementColumnNumber(svgElement)),svgDocEntry);
+	      if (width.contains("%") || height.contains("%")){
```
rdeltour: I think this test is too permissive: the spec says the values should be defined in **pixel** values, i.e. all other non-percent units should trigger the warning too.

I suggest we change the test by matching to a px-value regex (something like `\d+px|0`)

src/main/resources/com/adobe/epubcheck/messages/MessageBundle.properties
```diff
@@ -114,14 +114,16 @@ HTM_046=Fixed format item has no viewport defined.
...
-HTM_048_SUG=A viewBox declaration is required for fixed format documents.
+HTM_048=SVG ViewBox and width and height are missing on fixed format document.
+HTM_048_SUG=A viewBox declaration or width and height are required for fixed format documents.
 HTM_049=Html element does not have an xmlns set to 'http://www.w3.org/1999/xhtml'.
 HTM_049_SUG=Add xmlns="http://www.w3.org/1999/xhtml" to the html element.
 HTM_050=Found epub:type="pagebreak" attribute in content document.
 HTM_051=Found Microdata semantic enrichments but no RDFa. EDUPUB recommends using RDFa Lite.
 HTM_052=The property 'region-based' is only allowed on nav elements in Data Navigation Documents.
 HTM_053=Found an external file link (file://) in file: '%1$s'.
+HTM_054=Only 'width' and 'height' were found for an SVG element.  SVG elements should include a 'viewBox' property. File: '%1$s'
```
rdeltour: suggested rewording:
> The use of a 'viewBox' attribute is recommended on SVG elements.

rdeltour: btw, the "File:" information here should not be necessary, as this information is in the `EPUBLocation` object tied to the message…

src/main/resources/com/adobe/epubcheck/messages/MessageBundle.properties
```diff
@@ -114,14 +114,16 @@ HTM_046=Fixed format item has no viewport defined.
...
+HTM_055=Width and Height properties should be in pixel sizes for fixed format documents. File: '%1$s'
```
rdeltour: suggested rewording:
>The  'width' and 'height' attribute values should be set to absolute lengths using the pixel unit.

src/test/resources/30/expanded/valid/fxl-svg-width-height/EPUB/cover.svg
```diff
@@ -0,0 +1,262 @@
+<?xml version="1.0" encoding="utf-8"?>
```
rdeltour: A more minimal SVG would be helpful for easier readability and maintainability.

We could use sth like the following as a base document:

```svg
<?xml version="1.0" encoding="utf-8"?>
<svg version="1.1" xmlns="http://www.w3.org/2000/svg" width="100%" height="100%"
  viewBox="0 0 100 100">
  <desc>Test SVG</desc>

  <rect x="10" y="10" width="80" height="80" fill="none" stroke="blue" stroke-width="2"/>
</svg>
```
(in this specific test sample, we would remove the `viewBox` attribute).

rdeltour: Also, maybe the test file name 'fxl-svg-no-viewbox' would be more explicit?

rdeltour:
@mattgarrish could you please review my comments, and confirm if you agree with the suggested message rewording and check-logic? I added you as a reviewer.

src/main/java/com/adobe/epubcheck/ctc/EpubSVGCheck.java
mattgarrish: Right, I think you could just make the test \d+(px)? as pixels are the default unit, or at least that's what I get out of this para:

> For the outermost svg element, the SVG user agent must determine an initial viewport coordinate system and an initial user coordinate system such that the two coordinates systems are identical. The origin of both coordinate systems must be at the origin of the SVG viewport, and one unit in the initial coordinate system must equal one CSS 2.1 px ([CSS2], section 4.3.2) in the SVG viewport.

src/main/resources/com/adobe/epubcheck/messages/MessageBundle.properties
mattgarrish: +1 assuming you mean to replace all of what is there with your suggested rewording. It shouldn't be necessary to inform people what was found, so the more compact the message the better.

src/main/java/com/adobe/epubcheck/ctc/EpubSVGCheck.java
rdeltour: Yeah I'm not sure: CSS values says `For zero lengths the unit identifier is optional`; I don't think it's allowed to have unit-less length values other than "0".

src/main/resources/com/adobe/epubcheck/messages/MessageBundle.properties
mattgarrish: It sort of sounds a bit too techy. I'd maybe go a little simpler with:

> The values of the 'height' and 'width' attributes should be specified in pixels.

mattgarrish:
One thing that should also be tested (unless I missed it) is that using both height/width and viewbox is not recommended. After saying one or both method can be used, the content documents specification adds this:

> It is RECOMMENDED to use only the viewBox attribute to ensure proper rescaling of the SVG Content Document, as needed.

rdeltour:
Good point. Does this held for relative height/width though? My understanding is that it's an issue only with absolute height/width.

mattgarrish:
As far as I remember, the recommendation against using both viewbox and height/width was intended to apply no matter the units specified. @iherman or @rkwright would know more than I do about this, though.

src/main/java/com/adobe/epubcheck/ctc/EpubSVGCheck.java
mattgarrish: In the units section of SVG 1.1 it says:

> All coordinates and lengths in SVG can be specified with or without a unit identifier.
>
> When a coordinate or length value is a number without a unit identifier (e.g., "25"), then the given coordinate or length is assumed to be in user units (i.e., a value in the current user coordinate system).

I pulled that other quote out of SVG2, I think, as I followed the EPUB reference, but I don't think it's changed that the default user unit is pixel in these cases.

rdeltour: Right, I was just looking at SVG 2 and it says:

> If the width or height presentation attributes on the outermost svg element are in user units (i.e., no unit identifier has been provided), then the value is assumed to be equivalent to the same number of "px" units (see Units).

So let's go for `\d+(px)?`!

rdeltour:
I applied the suggested tweaks, and rebased+squashed+force-pushed the result.

Note:
- I **did not** implement the warning raised when both `viewBox` and `width`/`height` attributes are set, as it depends on the discussion happening in w3c/publ-epub-revision#1236
- I fixed a bug where all `svg` elements were checked, when only the outer svg element should be
- I minimized the SVG documents used in the `fxl-svg-*` tests.

@mattgarrish @GrayWolfMT if you have time to do one final quick review before merging, that would be great 😊

mattgarrish:
If I really wanted to nitpick, I'd say there should be an SVG message category, as the errors don't have anything to do with HTML, but that can be done another day.

Otherwise, lgtm.

rdeltour:
> If I really wanted to nitpick, I'd say there should be an SVG message category, as the errors don't have anything to do with HTML, but that can be done another day.

Yeah, I agree. Let's postpone that to after v4.2.0, along with the other message-code improvements (like the skipped numbers we talked about in another issue).

PR #993, opened 2019-03-07
Update rules for remote resources

### LINKED ISSUE #672: Error on SVG font files in pre-paginated EPUB rendition
If an EPUB rendition has the `rendition:layout` set to `pre-paginated`, SVG font files are getting flagged with this error:
`ERROR(HTM-048): test.epub/OEBPS//fonts/webfont.svg(x,xx): SVG ViewBox is missing on fixed format document.`

These files should not get validated as SVG content documents.

### LINKED ISSUE #869: Verification of use of remote resources
Scripts are allowed to reference remote resources, so whether a remote foreign resource listed in the manifest is in actual use or not may be complicated to check. It might make sense to only issue an informative message that no reference could be found, or emit nothing at all, if all resources are checked to see if they are in use.

Reference: https://w3c.github.io/publ-epub-revision/epub32/spec/epub-spec.html#sec-resource-locations-script

### LINKED ISSUE #871: Fonts may be located remotely
Fonts media types are now allowed to be hosted outside the container.

They must be marked as a "remote-resource" in the package document manifest.

Reference: https://w3c.github.io/publ-epub-revision/epub32/spec/epub-spec.html#sec-resource-locations-fonts

rdeltour:
### Feature changes:
- publication resources used in font declarations (in CSS documents,
  in inline CSS content of XHTML documents or SVG, or in SVG's
  deprecated `font-face-uri` element) MAY be remote resources.
- remote resource check is dependent on content parsing, so the check is
  moved to *after* the resource contents are checked, and only
  run when checking an entire publication.

### Fix:
- check that the item property "remote-resource" is properly set when
  required on items representing CSS documents

### Internal changes:
- Add a new `FONT` type to XRefChecker references
- Register as `XRefChecker` references of type `FONT` the paths found in
  CSS `@font-face` declaration and SVG `font-face-uri` elements.
- Add a new `getTypes(path)` public method in `XRefChecker`, to get the
  type(s) of reference(s) to a given publication resource
- Add a new `isRemote(String path)` util method to `PathUtil`
- `CSSHandler` constructor only receives a `ValidationContext`
- add tests

Fix #871, Fix #672

### Message changes:

- message `OPF-018` (`WARNING`) is only reported when a Content Doc is
  declared in the Package Document with the property `remote-resource`,
  and neither remote references nor scripts are found.
- new message `OPF-018b` (`USAGE`) reported when a Content Doc is
  declared in the Package Document with the property `remote-resource`,
  and no remote references are found but the Content Doc has scripted
  content.
- message `RSC-006` (`ERROR`) is only reported when a remote publication
  resource is declared in the Package Document with a type that doesn't
  specifically allow it to be located outside the container, and when no
  scripted content is found.
- new message `RSC-006b` (`USAGE`) is reported when a remote publication
  resource is declared in the Package Document with a type that doesn't
  specifically allow it to be located outside the container, and
  scripted content is found.

### Internal changes:

- new utility static method `OPFChecker.isScriptType(String)`
  tells if a media type is a [JavaScript MIME type essence match](https://mimesniff.spec.whatwg.org/#javascript-mime-type-essence-match)
- better identify scripted content: from event handler attributes, or
  script element with a type matching a JavaScript MIME type essence.
- add a couple tests and fix existing tests

Fix #869.


mattgarrish:
One problem I'm noticing is that the testing is very permissive once scripting is used.

If there is a script _anywhere_ in the EPUB, then you can reference remote resources from any HTML elements without complaint so long as you say the file contains remote resources.

I was expecting that epubcheck would only allow remote resources a pass if it couldn't find any reference to them in the content AND there was scripting.

mattgarrish:
By way of example, I put a reference to a remote jpeg into an img and iframe in separate files. epubcheck correctly complained that these were not allowed outside the container. But if I added a script to a third file then it stopped complaining.

rdeltour:
> If there is a script _anywhere_ in the EPUB, then you can reference remote resources from any HTML elements without complaint so long as you say the file contains remote resources.

Ah, I see what you mean. This sound like an unplanned shortcoming of the checking rule, I'll try to rework the logic. Good catch!

rdeltour:
I guess I misread this statement of the spec:

> Resources retrieved by scripts MAY be located outside the EPUB Container.

And assumed that as long as this was the case, no other checking was needed. The thing is, being retrieved by a script does not make it legit in other direct references (`img` element for instance). This is kinda implicit in the spec, but I'm wondering if there would be a way to make it more explicit.

mattgarrish:
> This is kinda implicit in the spec, but I'm wondering if there would be a way to make it more explicit.

I'd definitely like to bring xhr and fetch into the definition, as "retrieve" is vague.

rdeltour:
well, thank **you** for the good catch in your earlier review 😄

PR #996, opened 2019-03-08
chore: remove unused/unmaintained BookReporter python tool

tofi86:
The `BookReporter.py` Python tool has been introduced by B&N with 4.0 but has since never been touched. I suggest to remove it.


rdeltour:
Have you tried it recently? Does it still work or is it actually broken?

If it does work, then we could keep it. If it doesn't then yeah we could remove it… I'm not sure if anyone is actually using this tool (?), but it could possibly be maintained in a separate repo.

tofi86:
tbh I didn't try. I never used it.

moving it to another repo could be an idea.

rdeltour (approved):
OK to remove. We can go back in git history and extract to a new project if anyone demand these tools.

PR #998, opened 2019-03-10
Add doc-glossary to section

### LINKED ISSUE #997: allow doc-glossary on section elements
The doc-glossary role is currently not allowed anywhere by the schemas, but should be allowed where the other landmark roles are.

This is most likely the result of a bug in the 1.0 release, which identifies the role as a landmark but the example shows it on a `dl` element. Once this is corrected, we'll open bugs against the schemas, but we should allow for epub now given that this is already fixed in the editor's draft for 1.1.

mattgarrish:
This PR fixes #997 by allowing doc-glossary on section element.

Probably should include some tests, but we don't test any of the roles so figuring we can deal with that at a later time.


mattgarrish:
I'd adapt it here now, since this is the most relevant audience. From experience, they typically want to see a rec document before making a change, and all we have is a draft. I'll open an issue/PR against both the document and validator when I have a minute, though.

PR #1002, opened 2019-03-12
test: foreign resources unrendered don't need fallbacks

### LINKED ISSUE #868: Allow foreign resources without fallbacks when not rendered
Foreign resources are allowed in the EPUB container without fallbacks so long as they are not referenced from the spine or directly rendered in an HTML embedded content element or SVG image or foreignObject element.

This new provision means that resources can travel in the container without any obvious reference to them in the content. Any warning about unreferenced resources might need to become an informative message.

Reference: https://w3c.github.io/publ-epub-revision/epub32/spec/epub-spec.html#sec-foreign-restrictions

rdeltour:
These were already ignored; this PR just adds some tests.

Close #868

### DIFF
--- a/src/test/java/com/adobe/epubcheck/api/Epub30CheckExpandedTest.java
+++ b/src/test/java/com/adobe/epubcheck/api/Epub30CheckExpandedTest.java
@@ -118,6 +118,20 @@ public class Epub30CheckExpandedTest extends AbstractEpubCheckTest
   }
 
   @Test
+  public void testForeign_Unused()
+  {
+    // test that an unreferenced foreign resource MAY be included without fallback
+    testValidateDocument("valid/foreign-unused");
+  }
+  
+  @Test
+  public void testForeign_InLink()
+  {
+    // test that an foreign resource used in HTML link MAY be included without fallback
+    testValidateDocument("valid/foreign-in-link");
+  }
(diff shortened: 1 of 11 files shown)


mattgarrish:
We don't say that HTML `link` is exempt from fallback requirements - only when it has [the value `pronunciation`](https://w3c.github.io/publ-epub-revision/epub32/spec/epub-contentdocs.html#confreq-resources-cd-fallback-link). But if we haven't been checking, let's not start now.

I'm not sure why this restriction needs to be in place, to be honest, as linked resources aren't required to be used. Another for the CG.

rdeltour:
Yeah I agree the `link` checks are a bit iffy. In HTML, `link` can represent both a hyperlink or a link to an external resource depending on its `rel` attribute. So theoretically the spec would say that the former needs to be a CMT (in spine, even) and the latter could be foreign.

But then, some such hyperlinks definitely point to alternative versions of the document (“alternate “ or “canonical”) which could by essence be exempt of being a CMT, while some external resources are in fact used for rendering (style sheets).

So yeah the spec could be made clearer, and EPUBCheck could definitely do with a better logic. But since this was never reported nor specific to 3.2, let’s leave it to a later EPUBCheck version :)

PR #1005, opened 2019-03-13
fix: allow `epub:type` on all HTML elements

### LINKED ISSUE #986: epubcheck 4.2.0.-beta: meta element with attribute type from OPS namespace
epubcheck 4.2.0.-beta clains, that it is not correct to use the attribute type from the OPS namespace within the meta element:

Example:
"
&lt;html ...
xmlns:ops="http://www.idpf.org/2007/ops"
prefix="dct: http://purl.org/dc/terms/'
...
&gt;
...
&lt;meta name="creator" ops:type="dct:creator" property="dct:creator" content="Friedrich Schiller" /&gt;
"

Reports as:
"ERROR(RSC-005): SchillerndF.epub/Inhalt/andiefreude.xhtml(12,103): Error while parsing file: attribute "ops:type" not allowed here; ..."

EPUB 3.0.1 notes, that this OPS attribute is a global attribute ...
http://www.idpf.org/epub/301/spec/epub-contentdocs.html#sec-xhtml-content-type-attribute

rdeltour:
Fix #986

### DIFF
--- a/src/main/resources/com/adobe/epubcheck/schema/30/epub-nav-30.rnc
+++ b/src/main/resources/com/adobe/epubcheck/schema/30/epub-nav-30.rnc
@@ -7,9 +7,13 @@
    }
 
    ## "reset" the definition of nav.attrs to what it was
-   ## before epub:type was set to augment common.attrs
+   ## before epub:type was set to augment common.attrs.basic
    nav.attrs.noepubtype = 
-     ( common.attrs.basic
+     ( common.attrs.id?
+     & common.attrs.class?
+     & common.attrs.title?
+     & common.attrs.base?
+     & common.attrs.space?
      & common.attrs.i18n
      & common.attrs.present
      & common.attrs.other
--- a/src/main/resources/com/adobe/epubcheck/schema/30/mod/epub-type-attr.rnc
+++ b/src/main/resources/com/adobe/epubcheck/schema/30/mod/epub-type-attr.rnc
@@ -8,8 +8,6 @@
    ## common.attrs.other is augmented by other schemas.
    ## We therefore augment common.attrs (and also attrobute
    ## lists that are not based on common.attrs) 
-   common.attrs &= epub.type.attr?
-   a.attrs &= epub.type.attr?
-   area.attrs &= epub.type.attr?
+   common.attrs.basic &= epub.type.attr?
 
    epub.type.attr = attribute epub:type { datatype.properties }
--- a/src/test/java/com/adobe/epubcheck/ops/OPSCheckerTest.java
+++ b/src/test/java/com/adobe/epubcheck/ops/OPSCheckerTest.java
@@ -454,6 +454,13 @@ public class OPSCheckerTest
     testValidateDocument("xhtml/valid/epubtype.xhtml", "application/xhtml+xml",
         EPUBVersion.VERSION_3);
   }
+  
+  @Test
+  public void testEPUBTypeInHead()
+  {
+    testValidateDocument("xhtml/valid/epubtype-in-head.xhtml", "application/xhtml+xml",
+        EPUBVersion.VERSION_3);
+  }
 
   @Test
   public void testEPUBTypeFromReservedVocab()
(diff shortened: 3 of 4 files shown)


mattgarrish:
We might want to wait on a resolution to https://github.com/w3c/publ-epub-revision/issues/1241

rdeltour:
I think we need to check what previous versions of EPUBCheck did, we don’t want to introduce a de facto backward incompatibility.

But yeah, this one can wait for a CG decision; I wanted to have a PR ready just in case, but we can totally move it to the final release milestone.

mattgarrish:
Sure, it's just that if were going literally by what the spec says, epubcheck should allow the attribute but reject any values used in it, as it must express a subclass of the carrying element. As none of these are structural elements, they can't meet this requirement. It's not making a backwards incompatible change as much as enforcing the reality of a couple of confusing statements.

rdeltour:
So I just checked an 4.1.1 does allow `epub:type` on any elements, so yeah basically the new schemas are introducing a backward incompatibility **with how EPUBCheck previously worked** (regardless of what the spec says).

In fact, the current situation in 4.2.0-beta is a bit more complex that disallowing it on non-flow content. The fact that it's  reported as an error on `meta` element is a side effect of how we integrate it with the new HTML schemas, and was not really intentional. For instance, the current schemas allow `epub:type` on the `html` element or `title` element, which makes as much (or rather, little) sense as putting it on `meta` elements. So the current behavior is inconsistent at best 😊.

> if were going literally by what the spec says, epubcheck should allow the attribute but reject any values used in it, as it must express a subclass of the carrying element

Right, however we never did check the usage context (and probably never will)…

What I'm concerned about at this stage rejecting titles that were accepted in previous version of EPUBCheck and which –albeit not respecting the spec intention– did neither strongly violate the spec nor causing any harm to interoperability or the end users.

So all things considered I'd be rather favorable to merging this PR to fix the current inconsistency (both in behavior and with comparison to previous EPUBCheck versions); we can always implement a or better logic if/when the CG decides to be stricter. WDYT?

mattgarrish:
I'm not in favour of making epubcheck a determiner of how the specification is designed, so, no, I don't agree with this. One person trying to hack at metadata doesn't make a good case for introducing behaviour that is clearly wrong and entirely broken.

laudrain:
@rdeltour 

> What I'm concerned about at this stage rejecting titles that were accepted in previous versions of EPUBCheck

Me too !

rdeltour:
> I'm not in favour of making epubcheck a determiner of how the specification is designed
(…)
> One person trying to hack at metadata doesn't make a good case for introducing behaviour that is clearly wrong and entirely broken.

I totally agree with that!… but precisely: it's not about _introducing a new allowed usage_  (as previous EPUBCheck versions didn't catch that so **allowed it already**); it's rather about _not inconsistently and coincidentally **start to disallow it**_.

mattgarrish:
> not inconsistently and coincidentally **start to disallow it**

But this is what I strongly disagree with. There isn't an application of epub:type values that is valid, so how do you rationalize making _any_ change going forward to enforce a content restriction that has been missed or misapplied?

epubcheck has to allow everything that has been missed from now on, in other words, because that it would be inconsistent to apply the rules?

rdeltour:
> There isn't an application of epub:type values that is valid

and that's where _I_ disagree 😊. Currently the spec says about `epub:type`:
- > Global attribute. MAY be specified on all elements.
- > The inflected semantic MUST express a subclass of the semantic of the carrying element.

_semantics_ is defined in HTML as:
> Elements in the DOM represent things; that is, they have intrinsic meaning, also known as semantics.

You'll see that even non-flow elements do have semantics, as per this definition. E.g.

- > The head element represents a collection of metadata for the Document.
- > Links are a conceptual construct, created by a, area, form, and link elements, that represent a connection between two resources, one of which is the current Document.

So by a strict reading of the spec, I don't see what says that `epub:type` on a non-flow element is not conforming 😊

rdeltour:
To clarify: I won't argue that the spec should stay what it is; I'm absolutely fine if the CG wants to clarify the spec by sticking closer to the original intent. All I'm saying is that 1. given the current spec language, 2. and the previous behavior of EPUBCheck, 3. and the currently inconsistent behavior in 4.2.0-beta, then the cleanest (possibly intermediary) solution is to fall back to the previous status quo, i.e. allowing everywhere.

PR #1008, opened 2019-03-13
fix: parse CSS custom properties

### LINKED ISSUE #790: CSS custom properties error
TL;DR: when validating an EPUB file using CSS custom properties, a.k.a. CSS variables ([spec](https://www.w3.org/TR/css-variables/)), you get the following error: 

```
An error occurred while parsing the CSS: Token '-' not allowed here, expecting a property name.
```

It is my understanding this error occurs because epubcheck doesn’t support EPUB 3.1 yet but the spec is part of the [CSS Snapshot 2017](https://www.w3.org/TR/CSS/) (with a “rough interoperability” disclaimer though).

As the main [Readium CSS designer and maintainer](https://github.com/readium/readium-css/tree/master), I can tell a significant number of implementations are likely to ship with [CSS custom properties for user settings](https://gist.github.com/JayPanoz/c5bbf0bd7e53997d3a91d9c1be44a82f). There is also nothing we can do to prevent exposure to authors if needed.

rdeltour:
This change only fixes the `CssScanner` to parse CSS custom properties
as IDENT tokens.

Additional work may be needed to actually make the built-in CSS Grammar
aware of custom properties, but at this time this is not needed to
prevent EPUBCheck from reporting custom properties as errors.

Fix #790

### DIFF
--- a/src/main/java/org/idpf/epubcheck/util/css/CssScanner.java
+++ b/src/main/java/org/idpf/epubcheck/util/css/CssScanner.java
@@ -134,9 +134,10 @@ final class CssScanner
       {
         _urange();
       }
-      else if (NMSTART.matches(cur) || cur == '-'
-          && matches(next, NMSTART) || escapes.get(0).isPresent()
-          || (cur == '-' && escapes.get(1).isPresent()))
+      else if (NMSTART.matches(cur) 
+          || cur == '-'  && 
+             (matches(next, NMSTART) || reader.peek() == '-' || escapes.get(1).isPresent())
+          || escapes.get(0).isPresent())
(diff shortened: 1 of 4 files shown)


rdeltour:
> with the caveat that I'm far from an expert in css syntax parsing rules…

 yeah neither am I 😊 (although I do happen to go back to the CSS Syntax spec once in a while 🤓)… but all the tests pass (and the CSS parsing code, while a bit difficult to read because of the inherent complexity, seems to be quite well covered with tests. I also tested with @jaypanoz's ReadiumCSS test book which now passes with no errors or warnings!

PR #1012, opened 2019-03-15
feat: check image sources in `picture` and `srcset`

### LINKED ISSUE #781: picture element and image CMTs
Noting that HTML 5.1 adds the picture element, which allows multiple source images to be specified. This means that you can provide a fallback CMT image for a non-CMT image type without having to use manifest fallbacks.

epubcheck will need to account for the child source elements in determining whether a CMT is available.

This doesn't change fallbacks for images outside the picture element.

rdeltour:
Message changes:

- new `MED-007` (`ERROR`) message is reported when an image source
  references a foreign resource (unless on `source` elements which
  `type` attribute specify non-core media type).
- reworded `MED-003`; it now includes the resource path.

Internal changes:

- add a `SourceSet` class to represent an image source set, along with
  a parser folloing the "parsing a srcset attribute" algorithm from HTML.
  See https://html.spec.whatwg.org/#parsing-a-srcset-attribute
- add to new reference types to the `XRefChecker` class, `PICTURE_SOURCE`
  and `PICTURE_SOURCE_FOREIGN`, to identify image source references in
  `picture` elements.
- add tests, for both the `srcset` parser and the new fallback rules.

Fix #781


src/main/resources/com/adobe/epubcheck/messages/MessageBundle.properties
```diff
@@ -127,10 +127,11 @@ HTM_055=When no `viewBox` attribute is present, the values of the 'height' and '
 #media
 MED_001=Video poster must have core media image type.
 MED_002=%1$s element doesn't provide fallback.
-MED_003=Non-standard image resource of type %1$s found.
+MED_003=A manifest fallback must be provided for image resource '%1$s' of type '%2$s'.
 MED_004=Image file header may be corrupted.
 MED_005=Media Overlay audio reference %1$s to non-standard audio type %2$s found.
 MED_006=Some browsers do not support rendering SVG images which use a filename in the xlink:href property.
+MED_007=Within a 'picture' element, image sources must be Core Media Type Resources except for 'source' elements whose 'type' attribute is not a Core Media Type; found resource '%1$s' of foreign type '%2$s'.
```
mattgarrish: This is the only thing I'd maybe question. It's a mouthful to read this error. I wonder if we could shorten it to something like:

>  Foreign resources can only be referenced from a source element when its 'type' attribute is set.

rdeltour: Yeah I didn't know how to turn this. I tried other wordings but they didn't work. I like yours better however, it's simpler!

PR #1027, opened 2019-03-27
fix: report ZIP checks after the 'Validating…' message

### LINKED ISSUE #1025: Error appears out of order in validation sequence
```
ERROR(PKG-007): 2019_03_25T14_19_31_Z_9781091038721.epub/mimetype(-1,-1): Mimetype file should only contain the string 'application/epub+zip' and should not be compressed.
Validating using EPUB version 3.2 rules.
ERROR(NCX-001): 2019_03_25T14_19_31_Z_9781091038721.epub/EPUB/xhtml/raw/contents.ncx(-1,-1): NCX identifier ('8c4a714c-83d7-3683-e9bf-c704a38886af') does not match OPF identifier ('urn:uuid:8c4a714c-83d7-3683-e9bf-c704a38886af').
INFO(CSS-007): 2019_03_25T14_19_31_Z_9781091038721.epub/EPUB/xhtml/pdf9vtli/pages/pdf.css(2,28): Font-face reference EPUB/xhtml/pdf9vtli/pages/f1.woff refers to non-standard font type application-type/font-woff.
INFO(CSS-007): 2019_03_25T14_19_31_Z_9781091038721.epub/EPUB/xhtml/pdf9vtli/pages/pdf.css(3,28): Font-face reference EPUB/xhtml/pdf9vtli/pages/f2.woff refers to non-standard font type application-type/font-woff.
Check finished with errors
Messages: 0 fatals / 2 errors / 0 warnings / 2 infos
EPUBCheck completed
```
PKG-007 Errors seem to always appear before the "Validating" string when run in terminal. So far the only error I have seen that does this.

rdeltour:
Internal changes:

- move the zip header checks from the `EpubCheck` class to the
  `OCFChecker` class
- run the checks _after_ the "Validating using EPUB version…" message
- make sure the checks are run in any case even when the validation
  aborts before the "Validating using EPUB version…" message.

Test changes:

- add a CLI test case for the message ordering
- update the invalid mimetype test files to be based on the minimal EPUB
- remove a duplicate test case (`lorem-mimetype-2`)

Fix #1025


mattgarrish:
Slightly unrelated, but tied to the spec issue @dauwhe reported in https://github.com/w3c/publ-epub-revision/issues/1250 is that the epubcheck barfs if the mimetype has a carriage return at the end, but oddly is silent if you just whitespace pad the end of the filename. We should check no leading or trailing whitespace of any kind. I'll open a new issue for this.

rdeltour:
> Seems a bit unwieldy to have to know that checkzipheader needs to be run before any return, but can leave that kind of refactoring for another day.

Yeah I agree. This was the most straightforward way to both reorder the messages and make sure there was no behavior regression; this is definitely worth refactoring later, eg when we rework the reporting API.

rdeltour:
>  We should check no leading or trailing whitespace of any kind.

OK, I’ll add a commit to this PR.

PR #1030, opened 2019-03-29
fix: check trailing spaces in mimetype file

rdeltour:
See w3c/publ-epub-revision#1250

Apparently a few of our tests samples had trailing spaces in the `mimetype` file. This is probably due to copy/paste, but potentially previously-accepted EPUBs will now be rejected.

Do you think it's safe to change this now (i.e. after the RC) @mattgarrish, @dauwhe? Or should we raise that to the CG?


rdeltour:
@wareid would it be possible for you to get usage data on that issue? The question is: if EPUBCheck starts checking that the mimetype file has no trailing spaces, would it break existing EPUBs?

rdeltour:
@wareid 
> I don't think I would be able to get usage data on whether or not our mimetypes have trailing spaces, but I could test a batch of files against the change to see if it's broken/changed anything.

OK, thanks for answering 😊 The issue is that this isn't integrated yet, so the check isn't run by 4.2.0-RC. Would it work if I provide you with a custom build?

wareid:
Of course!

laudrain:
@wareid Am I wrong if I think it can be possible to make a difference between existing EPUB files that have been already ingested in vendors systems with previous checks, and new files that would be delivered by publishers to be distributed anew?
I mean that this EPUBCheck change would have no impact on old files, only on new files that would be better checked. Wich would be, IMO, a good move.

mattgarrish:
FWIW, I don't have an issue with being lenient with non-breaking "errors", but authors should know when their epubs have potentially serious issues in them. From what @dauwhe reported, there are reading systems that aren't able to open files when the mimetype has extra space/padding, so even if this means some fixing is necessary, it would be better to alert the author.

PR #1031, opened 2019-03-29
fix: remove restrictions on MathML annotation-xml

### LINKED ISSUE #1024: m:annotation-xml missing required attribute "name"
```
ERROR(RSC-005): rog_9781319055783_EM_endpaper.xhtml(561,1055): Error while parsing file: element "m:annotation-xml" missing required attribute "name"
```

Searching around, I was able to find what I thought was the solution to my problem (https://www.idpf.org/epub/30/spec/epub30-contentdocs.html#sec-xhtml-mathml-conf-content), but this has not come up in previous versions of epubcheck. 

After setting name to value contentequiv, I received another error: 

```
Error while parsing file: value of attribute "name" is invalid; must be equal to "alternate-representation"
```

At this point I was able to resolve the issue and validate the epub.

My question is two-fold: Should this be flagged at this level of severity? Could the error report elaborate more on suggested values of attribute name?

And for my own edification, where is this requirement coming from?

rdeltour:
- Content MathML is still restricted by the rules in the spec:
  https://w3c.github.io/publ-epub-revision/epub32/spec/epub-contentdocs.html#sec-xhtml-mathml-conf-content
- Presentation MathML is allowed
- SVG is allowed
- MathML descendants are allowed in HTML annotations

Fix #1024
See also w3c/publ-epub-revision#1212 and w3c/publ-epub-revision#1249


src/main/resources/com/adobe/epubcheck/schema/30/mod/epub-mathml3.rnc
```diff
@@ -96,6 +82,6 @@ epub.annotation-xml =
           }
         epub.att-name.mathml.presentation =
           attribute name {
-            string "contentequiv"
+            xsd:NCName
```
mattgarrish: name is only defined as expecting a [string value](https://www.w3.org/TR/MathML/chapter5.html#id.5.2.3.2).

rdeltour: yeah, I copied the definition from from the normative [MathML schemas](https://www.w3.org/TR/MathML/appendixa.html#parsing_name).
If I understand correctly, `name` represents a symbol in a [Content Dictionary](https://www.w3.org/TR/MathML/chapter4.html#contm.cds), and its syntax (along with the companion attribute `cd`) is defined in the [`csymbol` element](https://www.w3.org/TR/MathML/chapter4.html#contm.csymbol), where it seems to make sense for it to be an `NCName`…

Anyways, I suppose we can safely align with the MathML schemas there…

mattgarrish: Ya, it's weird that no mention is made why it can't have colons, but the csymbol definition points to the schema for its value definition and its ncname, so whatever...

PR #1037, opened 2019-04-14
Issue/975

GrayWolfMT:
Added checks for Link elements with "rel" attributes of "prev" and "next".  The checks test for the existence of the referenced objects within the ePub.  Checks ensure that the referenced files exist, that they are not remote resources, and whether they match the spine sequence.


rdeltour:
This PR needs to be rebased, reviewed, and tests ported to the new test framework. Assigning it to me for v4.2.6 (tentatively).

PR #1056, opened 2019-07-11
feat: revert the spine/toc nav order check to a WARNING

### LINKED ISSUE #1036: Sudden errors by different order of toc/spine by EPUBCheck v4.2.0-rc
We found EPUBCheck v4.2.0-rc will report errors when toc nav's order is different from spine order.
Now we found that wrong order was prohibited already in EPUB 3.0 but EPUBCheck v4.2.0-rc may be the first version to be checked properly.
(https://github.com/w3c/epubcheck/issues/888)

Unfortunately, we might already generate not a few EPUB files which have different toc order from spine because printed books have such toc order.

For example, our Manga content is structured like below.

[spine order]

- Chapter/Story 01
- Column 01
- Chapter/Story 02
- Column 02
...

[toc order]

- Chapter/Story 01
- Chapter/Story 02
...
- Column 01
- Column 02
...

Such different order of toc may be prepared because the value of chapter and column is different and such toc makes it easy for users to access to the prioritized chapters.
I'm not sure but other publishers in Japan may also have Manga contents with such toc of different order from reading order.

Although we will modify our EPUB files when we find to use such wrong order toc, I'd like to discuss the way to reconsider of this specification for reproducibility of printed books or I'd like to propose to change behavior of EPUBCheck to WARNING, or other messages, in order to keep availability of existing EPUB files.

As already reported mainly from Japan, errors by EPUBCheck will give critical impact because  it will prevent from delivering EPUB files to eBook stores in Japan.

I appreciate if you have any interest this topic.

rdeltour:
As agreed on the 2019-07-11 EPUB CG call:
https://www.w3.org/2019/07/11-epub3cg-minutes.html

Changes the behavior introduced in #999

Fixes #1036

### DIFF
--- a/src/main/java/com/adobe/epubcheck/messages/DefaultSeverities.java
+++ b/src/main/java/com/adobe/epubcheck/messages/DefaultSeverities.java
@@ -34,6 +34,8 @@ class DefaultSeverities implements Severities
     {
       return;
     }
+    // Info
+    severities.put(MessageId.INF_001, Severity.INFO);
 
     // Accessibility
     severities.put(MessageId.ACC_001, Severity.USAGE);
--- a/src/main/java/com/adobe/epubcheck/messages/DefaultSeverities.java
+++ b/src/main/java/com/adobe/epubcheck/messages/DefaultSeverities.java
@@ -154,7 +154,7 @@ class DefaultSeverities implements Severities
     severities.put(MessageId.NAV_008, Severity.USAGE);
     severities.put(MessageId.NAV_009, Severity.ERROR);
     severities.put(MessageId.NAV_010, Severity.ERROR);
-    severities.put(MessageId.NAV_011, Severity.ERROR);
+    severities.put(MessageId.NAV_011, Severity.WARNING);
 
     // NCX
     severities.put(MessageId.NCX_001, Severity.ERROR);
--- a/src/main/java/com/adobe/epubcheck/messages/MessageId.java
+++ b/src/main/java/com/adobe/epubcheck/messages/MessageId.java
@@ -28,6 +28,9 @@ import java.util.NoSuchElementException;
 
 public enum MessageId implements Comparable<MessageId>
 {
+  // General info messages
+  INF_001("INF-001"),
+  
   // Messages relating to accessibility
   ACC_001("ACC-001"),
   ACC_002("ACC-002"),
--- a/src/main/java/com/adobe/epubcheck/opf/XRefChecker.java
+++ b/src/main/java/com/adobe/epubcheck/opf/XRefChecker.java
@@ -548,6 +548,8 @@ public class XRefChecker
       report.message(MessageId.NAV_011,
           EPUBLocation.create(ref.source, ref.lineNumber, ref.columnNumber),
           (ref.type == Type.NAV_TOC_LINK) ? "toc" : "page-list", ref.value, orderContext);
+      report.message(MessageId.INF_001,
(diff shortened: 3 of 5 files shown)


mattgarrish:
What about the request to explain the severity? Can we just dump something in an info message?

mattgarrish:
For example, pump out a generic suggestion like: "This rule is under review and may change in a future release. See the issue tracker."?

It just feels really awkward to output stuff like this via epubcheck, but if we make something generic then perhaps it could just be attached/detached to issues if we end up in a similar place again??

rdeltour:
> What about the request to explain the severity? Can we just dump something in an info message?

ah right. I had started the PR before this idea was brought. I'll think about how to do this (we have "suggestion" messages, that take another line and are bound to a message, and that might be usable for that purpose).

rdeltour:
> For example, pump out a generic suggestion like: "This rule is under review and may change in a future release. See the issue tracker."?

We could add that to a new line, but then that might broke any tool that rely on regex parsing? (other messages consistently start with `INFO`, `USAGE`, `WARNING`, or `ERROR`…)

Or we can append that to the message text, but then the line would get quite long…

mattgarrish:
> We could add that to a new line, but then that might broke any tool that rely on regex parsing?

I have to admit to never turning on any verbosity, so I just assumed such things could come out with their error/warning but labeled as info or some such on a new line.

I have no problem putting a prefix in front of the message, if that's the issue. I just had no idea what it should be so didn't add to the suggested text.

rdeltour:
I ended up adding a new `INFO` message, which has the benefit of not needing to update the current messaging mechanism and is output by default.

The message will read:

```
WARNING(NAV-011): ./src/test/resources/30/expanded/invalid/nav-page-list-unordered-spine.epub/OPS/nav.xhtml(20,40): 'page-list' nav must be in reading order; link target 'OPS/content_002.xhtml#p2' is before the previous link's target in spine order.
INFO(INF-001): ./src/test/resources/30/expanded/invalid/nav-page-list-unordered-spine.epub/OPS/nav.xhtml(20,40): The previous rule is under review and may change to an 'ERROR' in a future release. See the discussion at https://github.com/w3c/publ-epub-revision/issues/1283
```

@bduga can you please confirm this is OK with what you suggested in the CG call?

bduga:
Looks good to me!

rdeltour:
@mattgarrish since your review is still marked as pending, can you please confirm you're OK too with the info message?

ShinyaTakami:
Let me post a kind of concern.

‘ERROR’ in the INFO message may not comfortable if the system using EPUBCheck extract ‘ERROR’ as text by regular expression from the result message.

rdeltour:
> ‘ERROR’ in the INFO message may not comfortable if the system using EPUBCheck extract ‘ERROR’ as text by regular expression from the result message.

I hadn't thought of that (and just merged the PR). The change did intend to **not** break existing regex-based output parsers, notably by relying on the existing message mechanism rather than just adding a new line. I would expect an output parser to use a regex checking that `ERROR` only happens at the beginning of the line…

Does your concern is grounded on an existing system?

ShinyaTakami:
Yes, if existing system will use /^ERROR/ or /ERROR\(.*\):/ for extraction, it's no problem but just use of /ERROR/ should be modified.

The objective to change ERROR to WARNING is to avoid error detection by eBook stores when accepting EPUBs, and then NOT using 'ERROR' word in the result message will be safer, I think.

Anyway, sorry for my too late reaction.

PR #1060, opened 2019-07-18
Release v4.2.2

rdeltour:


rdeltour:
@laudrain, @mattgarrish, @tofi86 I added you as (optional) reviewers in case you want to have a look at the change log or test the ZIP before I release.
I'll release tomorrow morning (July 19) if I hear nothing wrong before 😊

mattgarrish:
I thought I saw that it also results in the page-list ordering check also reverting to a warning. Worth noting, not that anyone seems to care about it? (We've sort of avoided that other part of the equation.)

Oddly, I noticed that the page list definition still has the old 3.0.1 wording; we missed that when the toc was rewritten.

rdeltour:
> I thought I saw that it also results in the page-list ordering check also reverting to a warning. Worth noting, not that anyone seems to care about it?

Good catch! I amended the release notes (but won't touch the changelog to not have to rebuild and retag the release just for that).

tofi86:
The change in 11747ab61f662b9a40afb04b7282b992b660a37e leaves one new message in Transifex untranslated.

But it's an INFO, so I'm good with ignoring it for the moment.

rdeltour:
> The change in [11747ab](https://github.com/w3c/epubcheck/commit/11747ab61f662b9a40afb04b7282b992b660a37e) leaves one new message in Transifex untranslated.
> 
> But it's an INFO, so I'm good with ignoring it for the moment.

Right, it was my assumption that a new INFO message would be OK. We can add translations as they come in patch releases.

PR #1081, opened 2019-09-17
Avoid usage of BufferedImage to avoid OutOfMemoryError

DocJM:
In some cases, 700KB .jpg can be about 8500x6000 pixels and BufferedImage consumes about 100MB.
So if you have a lot of such images you easily can get OOM. Do not sure why, but GC does not collect old BufferedImage during single validation.

### DIFF
--- a/src/main/java/com/adobe/epubcheck/bitmap/BitmapChecker.java
+++ b/src/main/java/com/adobe/epubcheck/bitmap/BitmapChecker.java
@@ -31,6 +31,7 @@ import java.util.Iterator;
 
 import javax.imageio.ImageIO;
 import javax.imageio.ImageReader;
+import javax.imageio.stream.FileImageInputStream;
 import javax.imageio.stream.ImageInputStream;
 
 import com.adobe.epubcheck.api.EPUBLocation;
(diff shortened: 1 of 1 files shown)


rdeltour:
Thanks for the PR @DocJM, and for the review @Ridman. 😊

I'm assigning this to myself to integrate in the next release.

rdeltour:
@DocJM, do you have a sample EPUB to help reproduce the error? Or how many JPEG of which size do you estimate are needed to reproduce an OOM?

src/main/java/com/adobe/epubcheck/bitmap/BitmapChecker.java
```diff
@@ -138,20 +139,14 @@ public ImageHeuristics getImageSizes(String imgFileName) throws
...
 
 
-      if (formatFromSuffix != null && formatFromSuffix.equals(formatFromInputStream)) {
+    if (formatFromSuffix != null && formatFromSuffix.equals(formatFromInputStream)) {
       // file format and file extension matches; read image file
-      
-      try {
-        BufferedImage image = ImageIO.read(tempFile);
-        if (image == null) {
-          report.message(MessageId.PKG_021, EPUBLocation.create(imgFileName));
```
rdeltour: Note: this potentially removes a case where `PKG_021` was raised before. This might never have been raised in practice (only one test is related to `PKG_021`, and this error message is raised at various code paths), but this need to be carefully reviewed.

PR #1084, opened 2019-10-28
Refectored Media Overlay Document tests to BDD format.

GrayWolfMT:
Refactored Media Overlay Documents for BDD format.


src/test/resources/epub3/mediaoverlays-document.feature
```diff
@@ -10,5 +10,82 @@ Feature: EPUB 3 Media Overlays Document
...
+  Scenario: SMIL Document with invalid clip-begin and clip-end formats
+    When checking document 'incorrect-clip-formats.smil'
+    Then the following errors are reported
+        | RSC-005 | value of attribute "clipBegin" is invalid |
+        | RSC-005 | value of attribute "clipEnd" is invalid |
+        | RSC-005 | value of attribute "clipBegin" is invalid |
+        | RSC-005 | value of attribute "clipEnd" is invalid |
+        | RSC-005 | value of attribute "clipBegin" is invalid |
+        | RSC-005 | value of attribute "clipEnd" is invalid |
+        | RSC-005 | value of attribute "clipBegin" is invalid |
```
rdeltour: to reduce the verbosity of this scenario, we could just count the `RSC-005` errors here:
```
Then error RSC-005 is reported 7 times
```

If we really wanted to check the message content, we could just check the last message, like:

```
And the last message contains 'value of attribute "clipBegin" is invalid'
```

(currently "last" above is not allowed, but I can amend the step definition if needed).

src/test/resources/epub3/mediaoverlays-document.feature
```diff
@@ -10,5 +10,82 @@ Feature: EPUB 3 Media Overlays Document
...
+    And no other errors or warnings are reported
+
+  Scenario: SMIL Document with undeclared property and prefix values
+    When checking document 'undeclared-attributes.smil'
+    Then the following errors are reported
+         | OPF-027 | Undefined property: 'not' |
+         | OPF-027 | Undefined property: 'exists' |
+         | OPF-028 | Undeclared prefix: 'foof' |
+         | OPF-027 | Undefined property: 'asfas' |
+         | OPF-027 | Undefined property: 'asfas' |
```
rdeltour: the test file looks overly complex, the same feature is tested multiple times. I would simplify the test file by keeping only the first `epub:type` attribute, with:
- an example of a well-defined property in the default prefix
- an example of an undefined property in the default prefix
- an example of an undefined property in declared prefix

```
epub:type="chapter unknown foaf:unknown"
```

PR #1086, opened 2019-11-05
converting the navcheckertest.java to cucumber tests

deborahgu:
The tests all pass. I'm not sure I renamed the files according to how you'd like them; it seems to follow the pattern and the document, but I'm not positive.


src/test/resources/epub3/navigation-document.feature
```diff
@@ -10,5 +10,95 @@ Feature: EPUB 3 Navigation Document
...
+      | RSC-005 | Heading elements must contain text |
+      | RSC-005 | Spans within nav elements must contain text |
+    And no other errors or warnings are reported
+
+  Scenario: Nav document has anchors that lack text
+    When checking document 'invalid-nav-empty-anchor.xhtml'
+    Then error RSC-005 is reported
+    And no other errors or warnings are reported
+
+  Scenario: Nav document has anchors that lack text
```
rdeltour: s/anchor/link/

+ the title should explain how this scenario differs from the previous one

src/test/resources/epub3/navigation-document.feature
```diff
@@ -10,5 +10,95 @@ Feature: EPUB 3 Navigation Document
...
+    When checking document 'invalid-nav-empty-anchor-spans.xhtml'
+    Then the following errors are reported
+      | RSC-005 | Anchors within nav elements must contain text |
+      | RSC-005 | Spans within nav elements must contain text |
```
rdeltour: I haven't looked at the details, but it seems weird that 2 errors are reported here… maybe the validation logic could be improved here?

src/test/resources/epub3/navigation-document.feature
```diff
@@ -10,5 +10,95 @@ Feature: EPUB 3 Navigation Document
...
+
+  Scenario: Nav document has landmarks that lack epub:type
+    When checking document 'invalid-nav-landmarks-missing-anchor-type.xhtml'
+    Then error RSC-005 is reported
+    And no other errors or warnings are reported
+
+  Scenario: Nav document has too many landmarks elements
+    When checking document 'invalid-nav-landmarks-multiple.xhtml'
+    Then error RSC-005 is reported
+    And no other errors or warnings are reported
```
rdeltour: In the first scenario above, "landmarks" represents a landmark target, in the second it represents a landmark nav. Maybe the titles can be more explicit. Eg the first could be along the lines of "Landmark links must have an `epub:type` attribute." and the second "Only one `landmarks` nav is allowed".

src/test/resources/epub3/navigation-document.feature
```diff
@@ -10,5 +10,95 @@ Feature: EPUB 3 Navigation Document
...
+
+  Scenario: Nav document has duplicate landmark elements
```
rdeltour: More explicit (taken from the spec): "The landmarks nav MUST NOT include multiple entries with the same epub:type value that reference the same resource, or fragment thereof."
(or something simpler but as descriptive)

src/test/resources/epub3/navigation-document.feature
```diff
@@ -10,5 +10,95 @@ Feature: EPUB 3 Navigation Document
...
+    When checking document 'invalid-nav-landmarks-duplicates.xhtml'
+    Then error RSC-005 is reported
+    And error RSC-005 is reported
```
rdeltour: the multiple errors could be explained (as suggested in a previous comment above)

src/test/resources/epub3/navigation-document.feature
```diff
@@ -10,5 +10,95 @@ Feature: EPUB 3 Navigation Document
...
+    And no other errors or warnings are reported
+
+  Scenario: Nav document has anchors that lack text
```
rdeltour: this title doesn't seem to describe the scenario, which is apparently about a two landmarks of the same epub:type, but pointing to different locations.

src/test/resources/epub3/navigation-document.feature
```diff
@@ -10,5 +10,95 @@ Feature: EPUB 3 Navigation Document
...
+    When checking document 'invalid-nav-multiple-pagelists.xhtml'
+    Then error RSC-005 is reported
+    And no other errors or warnings are reported
+
+  Scenario: Nav document has ordered list with no elements
+    When checking document 'invalid-nav-empty-ol.xhtml'
+    Then error RSC-005 is reported
+    And no other errors or warnings are reported
+
+  Scenario: Nav document lacks required headings
```
rdeltour: having no heading is fine for the `toc`, `page-list`, and `landmarks` nav; the title could be explicit that this is a requirement for other navs.

src/test/resources/epub3/navigation-document.feature
```diff
@@ -10,5 +10,95 @@ Feature: EPUB 3 Navigation Document
...
+    When checking document 'invalid-nav-missing-required-headings.xhtml'
+    Then error RSC-005 is reported
+    And no other errors or warnings are reported
+
+  Scenario: Nav document has anchors that lack text
+    When checking document 'invalid-nav-empty-anchor.xhtml'
+    Then error RSC-005 is reported
+    And no other errors or warnings are reported
```
rdeltour: this scenario seems to be a duplicate of a previous one.

src/test/resources/epub3/navigation-document.feature
```diff
@@ -10,5 +10,95 @@ Feature: EPUB 3 Navigation Document
         defined in the `navigation.feature` feature file.
 
   Background: 
-    Given EPUB test files located at '/epub3/files/epub/'
-    And EPUBCheck with default settings
+    Given EPUBCheck configured to check a Navigation Document
+    And test files located at '/epub3/files/navigation-document-xhtml/'
+
+  Scenario: Minimal nav document
```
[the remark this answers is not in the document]
mattgarrish: How about a common pattern for valid/invalid using something like verify/report.

So in this case, it might be: Verify a minimal navigation document.

While an invalid test case might then say something like "Report a navigation document without a toc".

(Sort of taking inspiration from schematron assertions, I suppose.)

PR #1099, opened 2020-01-23
[fix #1097] downgrade PKG-012 (filename contains non-ASCII) from WARNING to USAGE

### LINKED ISSUE #1097: PKG-012 is invalid?
EPUB specification [allows unicode filenames](http://idpf.org/epub/301/spec/epub-ocf.html#sec-container-filenames), however epubcheck produces a warning when validates EPUB with non-ASCII symbols in filenames:
```
WARNING(PKG-012): unicode.epub/OEBPS/test-é.xhtml(-1,-1): File name contains the following non-ascii characters: é. Consider changing the filename.
```
Is there a reason why PKG-012 warns in this case or PKG-012 should be removed?

slonopotamus:


rdeltour:
This sound like a reasonable request, although I'm not sure if the WARNING is due to widely-adopted practices, and if publishers rely on that.
Any idea @mattgarrish?

rdeltour:
Rebased and merged as f368ee513e44d9dac4a2cfb29bad3cf150f8de37

PR #1102, opened 2020-03-05
Conversion of OPS tests

mattgarrish:
Opening for feedback on the conversion. Still many tests to do.


rdeltour:
This is a duplicate of #1112

PR #1103, opened 2020-03-06
test: port tests from OPFCheckerTest.java

rdeltour:


src/test/resources/epub-indexes/indexes-package-document.feature
```diff
@@ -0,0 +1,53 @@
...
+    Then no errors or warnings are reported
+    
+  Scenario: An index collection may contain index-group child collections
+    When checking file 'index-collection-index-group-valid.opf'
+    Then no errors or warnings are reported
+    
+  Scenario: An index collection must only contain links to XHTML Content Documents
+    When checking file 'index-collection-resource-not-xhtml-error.opf'
+    Then error OPF-071 is reported
+    And the message contains "Index collections must only contain resources pointing to XHTML Content Documents"
```
mattgarrish: Is adding messages into the tests going to make their maintenance more time-consuming going forward? To change one, you'd now have to update the message itself plus every feature it might be raised in.

And more a question, but how do you decide when to include a message matching case?

rdeltour: Yeah I've been wondering about that TBH.

Regarding the brittleness of including the messages in the tests, I think it's generally OK as long as we don't have too many tests quoting the same message, which shouldn't be the case anyway. (ideally, we should only have a very few tests about a specific message). Also it's an easy refactoring with global search and replace. So I'm not too worried.

I tried to add messages when it helped with the understanding of the test (but to be fair, it does seem redundant in the snippet quoted above 😅).

The one case where the message matching test is absolutely useful is for the few generic errors that are associated to many different messages, like RSC-005 and RSC-017.

mattgarrish: >The one case where the message matching test is absolutely useful is for the few generic errors that are associated to many different messages, like RSC-005 and RSC-017.

Ya, I thought of this as I was falling asleep last night. (seriously, epubcheck issues will knock you right out! ;P )

Given all the different messages that might come out for RSC-005, for example, it definitely makes sense to be specific. There's always a chance that one error could be replaced by an entirely different one and the build process wouldn't care as the count would be the same.

src/test/resources/epub-indexes/indexes-package-document.feature
```diff
@@ -0,0 +1,53 @@
...
+  Note: 
+  - Tests related to EPUB Indexes package rules in a full EPUB publication
+    are defined in the `indexes-package.feature` feature file.
+  - Tests related to general conformance of Package Documents are defined
+    in the `epub3/package-document.feature` feature file.
+
+  Background: 
+    Given EPUB test files located at '/epub-indexes/files/package-document/'
+    And EPUBCheck configured to check a Package Document
+
```
mattgarrish: Don't you have to configure for the edupub profile, too?

rdeltour: No, I don't think so. As far as I understand and remember, these tests are for index collections that can be part of any EPUB, but not specifically for index-only EPUBs. So the rules are always triggered as part of the default check.

This may require some double-check, as my memories tend to fail me 😅. For one thing, the `OPFCheckerTests.java` code that was ported didn't specify the index profile.

mattgarrish: Ah, okay. I stole the file for the xhtml content documents tests and noticed it didn't have it but it is needed for each test.

rdeltour: Right, for XHTML the profile would need to be explicitly set to trigger the checks. When not sure, you can look at the Java code, if the profile appears explicitly in the test method.

rdeltour:
Note to self: need to port tests introduced in v4.2.3

PR #1127, opened 2020-04-16
add dockerfile and a guide to build using docker

novalagung:
this pr contains two things:

- add dockerfilte
- add a guide on how to build the project in docker

dockerizing epubcheck will make installation and execution easier, especially for anyone who doesn't have the java dependencies ready in their local computer

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+# build the epubcheck.jar file
+FROM maven:slim as builder
+
+WORKDIR /app
+COPY . .
+RUN mvn clean install
+
+# prepare runner for epubcheck.jar execution
+FROM openjdk:slim
+
+WORKDIR /app
+COPY --from=builder /app .
+RUN echo '#!/bin/bash\n java -jar /app/target/epubcheck.jar "${@:1}"\n' > entrypoint.sh
+RUN chmod +x entrypoint.sh
+
+ENV DATA_PATH=/data
+WORKDIR ${DATA_PATH}
+VOLUME ${DATA_PATH}
+
+ENTRYPOINT [ "/app/entrypoint.sh" ]
--- a/README.md
+++ b/README.md
@@ -32,6 +32,8 @@ Historical archives of discussions prior to October 2017 are stored at the old [
 
 ## Building EPUBCheck
 
+### Build from sources
+
 To build epubcheck from the sources you need Java Development Kit (JDK) 1.7 or above and [Apache Maven](http://maven.apache.org/) 3.0 or above installed.
 
 You will also need Python to be able to run the BookReporter and related tools.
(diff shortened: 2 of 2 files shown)


sardar7110 (approved):

sardar7110 (approved):

rdeltour:
Sorry for the late reaction on this PR 😅, and thanks @novalagung for the contribution 😊.

I'm not super familiar with docker best practices (is it a common practice to push docker files in git repos?), so I'll need to spend the time to read more or get external advice before merging this. 

Also, I noticed that you also pushed a Docker image for EPUBCheck on [Docker Hub](https://hub.docker.com/r/novalagung/epubcheck). This looks useful! With this image available on the hub, is it still useful to have a docker file in the repo?

Ideally we would need to setup a travis build step or GitHub action to automatically push new docker images for tagged builds…

I'm marking this PR as "needs review" for now!

novalagung:
hi @rdeltour thank you for the response.

> is it a common practice to push docker files in git repos?

Yes, it is. by including the docker file to the git repo, it'll be easier for anyone to perform build and the epubcheck source code. let's try to use this epubcheck as an example. 

Right after cloning the repo, to be able to run the project I have to ensure JDK is installed locally, maven as well, etc. it's not a big deal for anyone who already has everything ready in there environment. but for me who rarely code java, I have to set up a lot of things first.

By putting the docker file into the project, it'll make us easier to build and run the project. meaning I don't have to set up a lot of things, so the only docker is enough. I just need to run `docker build ...` command and the entire build process will happen. to run the executable (in a form of a docker image), the `docker run ...` command is used. The full guide is added to the readme file.

> Also, I noticed that you also pushed a Docker image for EPUBCheck on Docker Hub. This looks useful! With this image available on the hub, is it still useful to have a docker file in the repo?

Yes, I did push it to my docker hub. I and anyone can use that docker image to run epubcheck without the need to rebuild the project first, simply just run `docker pull ...` then `docker run ...`. at the start I was using the image on my own project, then I thought it'll be useful if I pushed it to docker hub so anyone can reuse it.

However, I would say that putting the docker file into the project is still mandatory because to be able to build the image we need the docker file. especially if there is a plan to enable automatic push to the Docker hub, the build process will require the docker file ready.

> Ideally we would need to setup a travis build step or GitHub action to automatically push new docker images for tagged builds…

I think so. It'll be better if epubcheck has an official docker hub image, and then the automatic update is enabled via Travis (or other CI/CD tools).

rdeltour:
Thanks for the info @novalagung, that's very useful! 👍
I'll try to have a look for the next milestone (no scheduled date yet).

rdeltour (approved):

rdeltour:
Merged in as d0ed6a49c3a778363666e2a406b7fa407959a962. Thanks again @novalagung!

Issue #220, opened 2013-10-15
IDPF encrypted fonts fail validation in 3.0RC1


rdeltour:
_From [ch...@hexforge.co.uk](https://code.google.com/u/102780199363534770000/) on November 08, 2012 18:26:33_

PROBLEM: If fonts are obfuscated using the IDPF method, a warning appears during validation:

Epubcheck Version 3.0-RC-1

Validating against EPUB version 2.0
WARNING: ../encrypted_sample.epub: resource OEBPS/Fonts/Avenir-Roman.otf cannot be decrypted

Check finished with warnings or errors METHOD: 1. Create an epub containing at least one IDPF-obfuscated font, with the appropriate META-INF/encryption.xml file
2. Validate with 3.0RC1 or 3.0b5, and observe the warning above.
3. Validate with 3.0b4, and no warning appears RESULT: An unexpected warning occurs EXPECTED: I expected the epub to validate without warnings. PLATFORM: The incorrect warning appears in 3.0b5 and 3.0RC1. It works correctly in 3.0b4.
Running on MacOS 10.7.5 ADDITIONAL INFO: I downloaded the source for RC1 and stepped through the encryption validation. It appears that com.adobe.epubcheck.ocf.EncryptionHandler constructs IDPFFontManglingFilter with a null parameter. This causes the filter to always return False from canDecrypt, resulting in the warning.

The attached example epub demonstrates the problem - it validates cleanly with 3.0b4, but fails with the decryption warning in 3.0b5 and 3.0RC1. For licensing reasons, the OTF file included is not a real font, but since epubcheck doesn't actually examine the font data, I believe it's valid for the purposes of reproducing the issue.

**Attachment:** [encrypted_sample.epub](http://code.google.com/p/epubcheck/issues/detail?id=220)

_Original issue: http://code.google.com/p/epubcheck/issues/detail?id=220_

PR #1136, opened 2020-05-01
feat: downgrade RSC-004 (cannot decrypt resource) to INFO

### LINKED ISSUE #1077: null pointer exception on valid ePub w/ obfuscated SVG
With an obfuscated SVG image, current ePubCheck exits with a null pointer exception.

Minimal Test case: https://librelamp.com/testCase.epub

If the SVG image is not obfuscated the same ePub validates.

If the same obfuscation method is used to obfuscate fonts there is no issue, it is only an issue when obfuscating an SVG.

ePub viewers display the test case, including the obfuscated SVG, just fine.

I still am not able to run ePubCheck myself because of a Java dependency issue but https://www.mobileread.com/forums/showpost.php?p=3887641&postcount=7 shows the error when run by Doitsu at mobileread forum

rdeltour:
Unless I'm missing something, the spec doesn’t say that a reading system MUST or SHOULD support EPUB encryption.

EPUBCheck currently cannot decrypt encrypted or obfuscated resources. Previously, it reported any attempt to read an encrypted resource as an error, RSC-004. Internally, the code declared obfuscated resources as readable, even if it cannot practically read it, to not report obfuscated fonts as errors.
This is problematic for resources that EPUBCheck will try to parse deeper down the validation workflow (like SVG). The resource is declared as readable, so EPUBCheck tries to parse it, and fails with a `NullPointerException` (issue #1077).

The inability to read encrypted or obfuscated resources is a  limitation of EPUBCheck. EPUBCheck should:
- report the inability to read the resource as an informative-only message
- do its best to carry on the validation

This PR tries to do that by downgrading the severity of RSC-004 to `USAGE`. It also updates the de-encryption filters to accurately report that they cannot read the content. Validation of encrypted or obfuscated resource will then abort early, before trying to parse the content.

Obfuscated fonts, which are likely the most common usage of obfuscation, will keep on not being reported as warnings (see issue #220). The only difference is that they will now trigger the RSC-004 usage message.

Fixes #1077


rdeltour:
Fixes #1077

mattgarrish:
I'm kind of torn on this one. Issuing a warning sounds like the right thing to do, not because of anything the specification says, but because it can lead to content slipping by without being thoroughly checked. The usage message isn't going to show up unless you turn it on, right?

It sort of seems like something the person validating should silence through a flag, so that it's a conscious decision to ignore the encrypted/obfuscated content and any problems that might exist in it.

rdeltour:
Maybe using `INFO` instead be appropriate then? INFO messages are displayed by the command line by default. In some way, the message is an information on EPUBCheck’s limitation.

rdeltour:
@mattgarrish I updated the PR.

Running it on the obfuscated SVG test will give the following output:

```
$ epubcheck --mode exp src/test/resources/30/expanded/valid/container-obfuscation-svg-valid
Validating using EPUB version 3.2 rules.
INFO(RSC-004): ./src/test/resources/30/expanded/valid/container-obfuscation-svg-valid.epub(-1,-1): File "EPUB/emoji.svg" could not be decrypted.
No errors or warnings detected.
Messages: 0 fatals / 0 errors / 0 warnings / 1 info

EPUBCheck completed
```

PR #1139, opened 2020-05-04
update translations for v4.2.3

rdeltour:
This PR contains the updated translations for v4.2.3. It can be updated with newer translations until v4.2.3 is published.

The updated files have been pulled with `./src/build/transifex-pull.sh --all`

@tofi86 :
- I manually reviewed the messages in the Transifex UI to delete the translations which only contained a copy of the English message (to revert the side effect introduced in #1132).
- the pull command post-processing gives me a sed error (`sed: RE error: illegal byte sequence`) for the `com/thaiopensource/relaxng/pattern/resources/Messages_fr.properties` properties. I haven't looked deeper, but it may result in a corrupted or wrongly-escaped property file. Rings any bell?


tofi86:
@rdeltour sorry for not beeing able to review this or answer your question :-( You still have problems with the tx pull?

rdeltour:
@tofi86 no worries 😊
Yes the tx pul wasn’t able to properly encode one properties file… for now I manually edited it, and will see if the problem persists in the next pull 😉

PR #1142, opened 2020-05-20
Allow doc-epigraph and doc-cover

### LINKED ISSUE #1019: ERROR RSC-005 thrown on role="doc-epigraph"
Used _role="doc-epigraph"_ in a section tag and I'm getting **ERROR RSC-005 : Error while parsing file: value of attribute "role" is invalid; **. I'm not seeing that this ARIA role has been deprecated or done away with.

mattgarrish:
This fixes #1019 and #1141 by overriding the section attrs definition to allow doc-epigraph.

I also discovered that doc-cover is not allowed on img, so it adds an override for that. It's not currently listed in the ARIA in HTML document.

I couldn't find any more elegant way to add these semantics, as there's no way to inject them into the allowed roles as the lists are element-specific.

Both these patches should only be temporary, though, as the HTML schemas should eventually be updated to allow them.

### DIFF
--- a/src/main/resources/com/adobe/epubcheck/schema/30/mod/epub-xhtml.rnc
+++ b/src/main/resources/com/adobe/epubcheck/schema/30/mod/epub-xhtml.rnc
@@ -85,6 +85,7 @@ include "./html5/block.rnc"
 include "./html5/sectional.rnc"
 
 include "./html5/structural.rnc" {
+	# override to add doc-epigraph to section - remove when html schemas are updated
 	section.attrs =
 		(	common.attrs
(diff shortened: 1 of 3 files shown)


mattgarrish:
FYI, I found the removal of doc-cover came from this issue: https://github.com/w3c/html-aria/issues/230

I've opened a new issue to have it added back, which is why this was not an error in the last version of epubcheck but is now: https://github.com/w3c/html-aria/issues/230

mattgarrish:
I've opened an issue to have these added to the formal schemas: https://github.com/validator/validator/pull/966

PR #1143, opened 2020-05-21
Add DPUB-ARIA role tests

mattgarrish:
Adds tests for each element in the ARIA in HTML doc that has a restricted set of allowed DPUB-ARIA roles.

Setting this up as a separate PR from #1112 as the test for doc-epigraph will fail until merged with the new code base.

### DIFF
--- a/src/test/resources/epub3/files/content-document-xhtml/aria-roles-li-valid.xhtml
+++ b/src/test/resources/epub3/files/content-document-xhtml/aria-roles-li-valid.xhtml
@@ -6,13 +6,13 @@
 	</head>
 	<body>
 		<h1>Test</h1>
-		<section>
+		<section role="doc-endnotes">
 			<h2>Notes</h2>
 			<ol>
 				<li role="doc-endnote">…</li>
 			</ol>
 		</section>
-		<section>
+		<section role="doc-bibliography">
 			<h2>Bibliography</h2>
 			<ul>
 				<li role="doc-biblioentry">…</li>
--- a/src/test/resources/epub3/files/content-document-xhtml/aria-roles-li-valid.xhtml
+++ b/src/test/resources/epub3/files/content-document-xhtml/aria-roles-li-valid.xhtml
@@ -2,7 +2,7 @@
 <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 	<head>
 		<meta charset="utf-8" />
-		<title>Allow doc-endnote on list items</title>
+		<title>DPUB-ARIA roles allowed on li</title>
 	</head>
 	<body>
 		<h1>Test</h1>
@@ -13,9 +13,9 @@
 			</ol>
 		</section>
 		<section>
-			<h2>Notes</h2>
+			<h2>Bibliography</h2>
 			<ul>
-				<li role="doc-endnote">…</li>
+				<li role="doc-biblioentry">…</li>
 			</ul>
 		</section>
 	</body>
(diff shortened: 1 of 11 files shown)


src/test/resources/epub3/files/content-document-xhtml/aria-roles-li-valid.xhtml
```diff
@@ -13,9 +13,9 @@
 			</ol>
 		</section>
 		<section>
-			<h2>Notes</h2>
+			<h2>Bibliography</h2>
 			<ul>
-				<li role="doc-endnote">…</li>
+				<li role="doc-biblioentry">…</li>
```
rdeltour, line 18: Shouldn't this appear within the (required) context role `doc-bibliography`?
Also, why remove the `doc-endnote` test?

src/test/resources/epub3/files/content-document-xhtml/aria-roles-section-valid.xhtml
```diff
@@ -0,0 +1,93 @@
+<!DOCTYPE html>	 
+<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
+	<head>
+		<meta charset="utf-8" />
+		<title>DPUB-ARIA roles allowed on section</title>
```
rdeltour, line 5: some roles have required owned elements, so the test case is only "valid" since the schema do not check these relationships entirely.

I suggest we add the required context / owned elements everywhere needed to make the test cases theoretically valid (and consequently more future-proof).

rdeltour:
Also: note I rebased the branch to the latest `test/cleanup` and force pushed, so beware if you try to push/pull in your working copy 😉

src/test/resources/epub3/files/content-document-xhtml/aria-roles-li-valid.xhtml
mattgarrish: That's just some weirdness in the diff because I started with an existing file that had doc-endnote in it. The doc-endnote role is in the next section; it hasn't been removed.

But in response to this and your other comment, I questioned even having these two roles. They're the only required children roles and they're also overwhelmingly likely to be deprecated in the now soon-to-start 1.1 revision. Once that happens, their section roles will exist and not require them and these two will need to be moved to invalid tests.

Should we just strip these list item tests for now and put them formally into the invalid set once the revision gets going?

rdeltour: > That's just some weirdness in the diff because I started with an existing file that had doc-endnote in it. The doc-endnote role is in the next section; it hasn't been removed.

Oh, sorry I missed that 👍



> Should we just strip these list item tests for now and put them formally into the invalid set once the revision gets going?

Yeah, I don't know; it also depends on how soon this will happen.
I think I'm more  in favor of updating the test files to represent the spec as it currently is, and then fix it back later…

mattgarrish: > it also depends on how soon this will happen

Any day now is my understanding. The ARIA group just has to form a subgroup and we'll be underway.

I can add roles onto the parent sections to make them theoretically valid, but I'd then leave the section tests alone so we cover both scenarios. That way we can drop this whole file into the invalid tests when the time comes and nothing else needs to change.

rdeltour: OK, SGTM. Don't bother with making the changes, I can do it.

PR #1169, opened 2020-07-29
Test and check playback classes do not have refines attribute

mattgarrish:
A couple of minor schema updates and tests for the ["must not" in s. 3.4](https://www.w3.org/publishing/epub/epub-mediaoverlays.html#sec-docs-assoc-style).

### DIFF
--- a/src/main/resources/com/adobe/epubcheck/schema/30/package-30.sch
+++ b/src/main/resources/com/adobe/epubcheck/schema/30/package-30.sch
@@ -99,36 +99,6 @@
         </rule>
     </pattern>
 
-    <pattern id="opf.media.overlay">
-        <rule context="opf:item[@media-overlay]">
-            <let name="ref" value="./normalize-space(@media-overlay)"/>
-            <let name="item" value="//opf:manifest/opf:item[normalize-space(@id) = $ref]"/>
-        	<let name="item-media-type" value="normalize-space($item/@media-type)"/>
-            <assert test="$item-media-type = 'application/smil+xml'">media overlay items must be of
-                the "application/smil+xml" type (given type was "<value-of select="$item-media-type"
-                />")</assert>
-        </rule>
-    </pattern>
-
-    <pattern id="opf.media.overlay.metadata.global">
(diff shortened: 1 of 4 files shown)


src/main/resources/com/adobe/epubcheck/schema/30/package-30.sch
```diff
@@ -302,6 +272,52 @@
...
+			<let name="mo-item-id" value="$mo-item/normalize-space(@id)"/>
+			<let name="mo-item-uri" value="concat('#', $mo-item-id)"/>
+			<assert test="//opf:meta[normalize-space(@property)='media:duration' and normalize-space(@refines) = $mo-item-uri ]">item
+				media:duration meta element not set (expecting: meta property='media:duration'
+				refines='<value-of select="$mo-item-uri"/>')</assert>
+		</rule>
+	</pattern>
+	
+	<pattern id="opf.media.overlay.metadata.active-class">
+		<rule context="//opf:meta[normalize-space(@property)='media:active-class']">
```
rdeltour, line 308: probably nitpicking (I'm not sure there's any performance impact), but the leading double-slash can be removed 😊

src/main/resources/com/adobe/epubcheck/schema/30/package-30.sch
```diff
@@ -302,6 +272,52 @@
...
+			<report test="@refines"> @refines must not be used with the media:active-class property</report>
+		</rule>
+	</pattern>
+	
+	<pattern id="opf.media.overlay.metadata.playback-active-class">
+		<rule context="//opf:meta[normalize-space(@property)='media:playback-active-class']">
```
rdeltour, line 314: same comment as above

src/main/resources/com/adobe/epubcheck/schema/30/package-30.sch
mattgarrish: Ya, I imagine it's probably optimized out when the schema is compiled, but no need to add clutter.

rdeltour (approved):

PR #1171, opened 2020-08-03
match epub:type checking in overlays documents to content documents

mattgarrish:
Processing of epub:type is supposed to be constrained identically to content documents, which means as of 3.2 unknown values without a prefix are usage messages.

This updates the processing to match what is in place for content documents by aggregating unchecked_vocab and then iterating the value(s) after to issue usage message(s).


src/test/resources/epub3/mediaoverlays-smil-document.feature
```diff
@@ -93,8 +93,7 @@ Feature: EPUB 3 ▸ Media Overlays ▸ SMIL Document Checks
     And the message contains 'Undeclared prefix: "my"'
     And no other errors or warnings are reported
 
-  Scenario: Report an unknown epub:type property in the default vocabulary
-    When checking document 'epubtype-unknown-error.smil'
-    Then error OPF-027 is reported
-    And no other errors or warnings are reported
+  Scenario: Allow unknown epub:type properties in the default vocabulary
+    When checking document 'epubtype-unknown-valid.smil'
+    Then no errors or warnings are reported
```
rdeltour, line 98: The test could be improved by checking that the right USAGE message is reported:
```gherkin
Given the reporting level set to usage
When checking document 'epubtype-unknown-usage.smil'
Then usage OPF-088 is reported
```

mattgarrish: Good point! I was running without usage messages on.

rdeltour (approved):

PR #1173, opened 2020-08-06
Improve handling of audio clip times in media overlay documents

mattgarrish:
This PR adds better support for the processing of the `clipBegin`/`clipEnd` attributes as follows:

- Adds the DAISY SmilClock class for processing SMIL clock values
- Moves the test of `clipBegin`/`clipEnd` equality into Java code so it isn't a simple string comparison
- Adds a test that `clipBegin` is not temporally after `clipEnd`
- Adds and updates associated tests

### DIFF
--- a/src/main/java/com/adobe/epubcheck/messages/DefaultSeverities.java
+++ b/src/main/java/com/adobe/epubcheck/messages/DefaultSeverities.java
@@ -144,6 +144,8 @@ class DefaultSeverities implements Severities
     severities.put(MessageId.MED_005, Severity.ERROR);
     severities.put(MessageId.MED_006, Severity.USAGE);
     severities.put(MessageId.MED_007, Severity.ERROR);
+    severities.put(MessageId.MED_008, Severity.ERROR);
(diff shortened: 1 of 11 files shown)


mattgarrish:
> keep it in the `org.daisy.util.xml` package or take ownership of it and put it in an EPUBCheck package

Ya, I struggled with this. My first thought was just to put it in org.adobe.epubcheck, but then I started to question that since it was developed by daisy.

But I suppose all LGPL requires is that the copyright notice remain and that others can modify any changes we make. It doesn't say we can't change the source to place the class in another namespace.

So I'm fine changing the namespace and just leaving the daisy copyright. Would you recommend putting the class in epubcheck.util or .overlay or someplace else?

mattgarrish:
Wait, sorry, I thought you were just wanting to put it under org.adobe.epubcheck. If we want to take full ownership, that's definitely more complicated. We'd at least need a formal nod from daisy (maybe not everyone who contributed since its not copyright to individuals), as we'd be breaking the terms of the original license.

PR #1174, opened 2020-08-06
Verify media:duration property values are valid smil clock values

mattgarrish:
The PR adds a schematron test to check that media:duration values are valid smil clock values based on the regex used in the media overlays rnc to verify clipBegin and clipEnd.

### DIFF
--- a/src/main/resources/com/adobe/epubcheck/schema/30/package-30.sch
+++ b/src/main/resources/com/adobe/epubcheck/schema/30/package-30.sch
@@ -128,6 +128,14 @@
                     refines='<value-of select="$mo-item-uri"/>')</assert>
         </rule>
     </pattern>
+	
+	<pattern id="opf.duration.metadata.item">
+		<rule context="//opf:meta[normalize-space(@property)='media:duration']">
+			<assert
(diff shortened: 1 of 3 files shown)


src/main/resources/com/adobe/epubcheck/schema/30/package-30.sch
```diff
@@ -128,6 +128,14 @@
                     refines='<value-of select="$mo-item-uri"/>')</assert>
         </rule>
     </pattern>
+	
+	<pattern id="opf.duration.metadata.item">
+		<rule context="//opf:meta[normalize-space(@property)='media:duration']">
```
rdeltour, line 133: the leading double-slash can be removed

src/test/resources/epub3/mediaoverlays-package-document.feature
```diff
@@ -43,7 +43,14 @@ Feature: EPUB 3 ▸ Media Overlays ▸ Package Document Checks
...
   Scenario: the 'media:duration' property can be expressed as a timecount value
     When checking file 'mediaoverlays-duration-timecount-valid.opf'
-    Then no errors or warnings are reported
\ No newline at end of file
+    Then no errors or warnings are reported
+  
+  Scenario: Repoort 'media:duration' properties with non-clock values
+    When checking file 'mediaoverlays-duration-clock-values-error.opf'
+    Then error RSC-005 is reported 3 times
+    And the message contains "must be a valid SMIL3 clock value"
```
rdeltour, line 54: note that this message assertion will only apply to the last RSC-005 error. If you want to enforce it to all 3, you need to use the table syntax (see step `the following errors are reported:`).

rdeltour (approved):

PR #1175, opened 2020-08-06
Check remote-resources property is set for media overlay documents

mattgarrish:
This PR makes the following changes:

- adds a check on audio start tags to flag when there are remote resource references
- adds a check on the end of the smil document to verify that remote-resources is specified on the package item declaration

Also adds a test to report an error when the property is not set.

### DIFF
--- a/src/test/resources/epub3/package-publication.feature
+++ b/src/test/resources/epub3/package-publication.feature
@@ -174,6 +174,11 @@ Feature: EPUB 3 ▸ Packages ▸ Full Publication Checks
     Then warning OPF-018 is reported
     And no other errors or warnings are reported
 
+  Scenario: Report a media overlay document with remote resources but missing the `remote-resources` property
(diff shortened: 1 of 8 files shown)


src/test/resources/epub3/package-publication.feature
```diff
@@ -174,6 +174,11 @@ Feature: EPUB 3 ▸ Packages ▸ Full Publication Checks
     Then warning OPF-018 is reported
     And no other errors or warnings are reported
 
+  Scenario: Report a media overlay document with remote resources but missing the `remote-resources` property
+    When checking EPUB 'package-manifest-prop-remote-resource-overlays-error'
```
rdeltour, line 178: to better align with other tests for `remote-resource` property, I suggest the test file is renamed `resources-remote-audio-in-overlays-missing-property-error` 😅

PR #1176, opened 2020-08-07
Check media types of package items that declare media-overlay attribute

mattgarrish:
The `media-overlay` attribute is only allowed on manifest items for EPUB content documents. This PR adds a schematron check of the media type of each item to ensure this is true.


src/test/resources/epub3/mediaoverlays-package-document.feature
```diff
@@ -23,6 +23,12 @@ Feature: EPUB 3 ▸ Media Overlays ▸ Package Document Checks
     And the message contains 'must be of the "application/smil+xml" type'
     And no other errors or warnings are reported
 
+  Scenario: Report use of the media-overlay attribute on non-EPUB Content Documents
+    When checking file 'mediaoverlays-mo-attr-error.opf'
```
rdeltour, line 27: I would suggest naming the test file `mediaoverlays-for-non-content-document-error` would be a little more explicit, but it's nitpicking 😊

Issue #1225, opened 2021-03-14
Media overlay element ordering does not have to match content


mattgarrish:
The requirement for par and seq elements to match the default reading order has been moved to the accessibility specification. Epubcheck shouldn't emit errors now if there's a mismatch.

https://github.com/w3c/epub-specs/issues/1458

rdeltour:
@murata2makoto said in #1244
> I understand that this newly-added check ensures what has been required by the EPUB3 MO specification. This requirement is stated as "The ordering of these elements MUST match the default reading order of the EPUB Content Document." But there appear to be a number of Japanese DAISY publications that rely on some mismatch between the reading order and the occurrence order in MO documents.
>
> We understand that some EPUB software such as the DAISY pipeline does not work when such mismatch occurs in an EPUB or DAISY publication. But Japanese DAISY people have used EPUB or DAISY readers that can live with this error.
>
> This issue is quite similar to the mismatch between the navigation document and the spine document. We thus propose that such mismatch be treated as a warning rather than an error.

rdeltour:
I think it's reasonable to preemptively demote MED_015 to a warning in the next maintenance release, if it happens before fully implementing EPUB 3.3.

rdeltour:
EPUBCheck v4.2.6 now reports `MED-015` as a `USAGE`. I'm keeping this issue open until we decide if we want to totally prune the feature and related code when we implement EPUB 3.3.

rdeltour:
Closing this. `MED-015` is still reported as a `USAGE` in v5.0.0. I think it may be helpful to some, so let's not remove that check entirely.

PR #1259, opened 2021-06-29
feat: do not report Media Overlays ordering mismatch

rdeltour:
This PR demotes MED-015 to a USAGE. EPUB 3.3 will no longer required that Media Overlays are in content order. 

This proactively implements #1225, and goes a little further to what was proposed in the discussion by using `USAGE` instead of a `WARNING`.


mattgarrish (approved):
Usage is right, as the spec doesn't say anything about this issue. It's only a recommendation of the accessibility spec. We'll have to figure out how to check that in Ace separately.

PR #1286, opened 2021-10-22
fixed description location: help_text japanese

mac-kn:


rdeltour:
I applied that change on our Transifex space (Transifex is the service we use to maintain translations). Thanks for the contribution, @mac-kn!

PR #1296, opened 2022-01-18
Update message about rule review to include rule id

### LINKED ISSUE #1295: INF-001 message is ambiguous in the JSON output
If you get the message INF-001, it begins: "The previous rule is under review..." which is fine in the STDOUT output, but in the JSON format, the messages aren't in the same order, so it's hard to identify which issue might be reviewed.  Is it possible to include the issue ID instead of "The previous rule" so that it can be related to the original issue regardless of message output order?

icklecows:
Updates the message about a rule being under review to include the rule ID.  This resolves the issue of the message being ambiguous when using json output, where the order is not the same as in the default format.

Resolves #1295


rdeltour:
`INF-001` is no longer in user, but I integrated the suggested message tweak for any later use. Thanks @icklecows!

PR #1324, opened 2022-03-26
Do not url decode relative references

eric-wieser:
Currently, `<item href="difficult_filename_%23.png">` is url-decoded to `difficult_filename_#.png` and then interpreted as a url fragment, which feels like incorrect behavior.

It's unlikely this change is in the right place, so I'm making this change mainly to see what CI says (I hope there's CI that runs on PRs...)


rdeltour:
This no longer happens since we rely on a standard-compliant URL parsing library (in #1396). I'm closing this PR (but thanks for the contribution @eric-wieser! 👍)

PR #1340, opened 2022-06-25
Correct link to #message-codes

doug-kline:
The on page link for message code was incorrect.  Had #messages-reference and should be #message-codes l:200


rdeltour:
Thanks @doug-kline. I also fixed this in the website source, so that the good link will be produced when re-generating the site.

PR #1360, opened 2022-12-01
feat: add new Core Media Types (ECMAScript, OPUS, WebP)

### LINKED ISSUE #1189: Changes to core media types in 3.3
The current draft of EPUB 3.3 supports the following new core media types:

- Opus (audio/opus) as a core audio type. See https://github.com/w3c/epub-specs/issues/645
- WebP (image/webp) as a core image type. See https://github.com/w3c/epub-specs/issues/1344

### LINKED ISSUE #1249: Allow application/ecmascript as core media type
Scripts can now also be identified by the media type `application/ecmascript` in package document manifest `item` elements.

See https://github.com/w3c/epub-specs/issues/1353

rdeltour:
Added the following Core Media Types:
- "audio/opus"
- "image/webp"
- "application/ecmascript"

Also added some missing tests for some older core media types.

Fix #1249, Fix #1189


src/test/resources/epub3/content-publication.feature
```diff
@@ -198,6 +198,9 @@ Feature: EPUB 3 ▸ Content Documents ▸ Full Publication Checks
     Then error OPF-013 is reported
     And no other errors or warnings are reported
 
+  #####  script
+  
+  #//TODO verify script core media types
```
mattgarrish, line 203: Are you thinking of adding tests on the media type for script src references here?

(I don't think it matters, spec-wise, what you use in the content so long as one of the three JS types is declared is declared in the manifest, but maybe could be a usage note.)

mattgarrish (approved):

src/test/resources/epub3/content-publication.feature
mattgarrish: Or are you thinking of putting text/vbscript or some junk in to make sure it's checked to be a CMT?

mattgarrish (approved):

src/test/resources/epub3/content-publication.feature
rdeltour: the idea was to add tests for other HTML-allowed Javascript MIME types, like `application/x-javascript`, `text/ecmascript`, `text/jscript`, etc…

If the script is declared as such in the manifest, a warning/error should be raised that it is not a CMT.

PR #1361, opened 2022-12-01
feat: update checking of the Package Document 'refines' attribute

### LINKED ISSUE #1226: Check that refines attribute references publication resources by manifest item id
It is now formally recommended that the refines attribute use fragment identifiers to reference resources listed in the manifest.

https://github.com/w3c/epub-specs/issues/1361

rdeltour:
- `refines` MUST be a relative URL
- `refines` SHOULD refer to a Publication Resource by ID
- `refines` MUST point to an existing ID when it is a fragment-only URL

Fix #1226

### DIFF
--- a/src/main/resources/com/adobe/epubcheck/schema/30/package-30.sch
+++ b/src/main/resources/com/adobe/epubcheck/schema/30/package-30.sch
@@ -51,7 +51,22 @@
     <!-- Metadata checks -->
 
     <pattern id="opf.refines.relative">
-        <rule context="*[@refines and starts-with(normalize-space(@refines),'#')][not(ancestor::opf:collection)]">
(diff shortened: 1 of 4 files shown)


mattgarrish (approved):

mattgarrish:
If collections weren't mostly dead, I'd suggest checking that the rules don't apply to refines within one, but ultra-low priority at this point.

rdeltour:
> If collections weren't mostly dead, I'd suggest checking that the rules don't apply to refines within one, but ultra-low priority at this point.

you mean in the tests?
the implementation itself is correct as far as I can see, from the Schematron context `*[not(ancestor::opf:collection)][@refines]`.

PR #1365, opened 2022-12-01
feat: make SVG content conformance checking informative-only

rdeltour:
EPUB 3.3 no longer requires that SVG content conforms to SVG content model requirements,
only that they are well-formed, that ID are uniques, and that they respect some additional
EPUB-specific requirements.
    
This PR:
      * introduces a new permissive RelaxNG schema for SVG, checking only the EPUB-specific
        requirements on the `title` and `foreignObject` content model
        see w3c/epub-specs#1323
      * removes checks on the value of the `requiredExtensions` attribute of `foreignObject`
        see w3c/epub-specs#1087
      * adapts the main XHTML to SVG schema driver to the new permissive SVG schema
      * adds various tests for EPUB-specific requirements

The PR also introduces two new codes, `RSC-024` and `RSC-025`, to report informative schema checks.
    
An `XMLValidator` instance can now be instantiated with a boolean flag to tell if the unerlying schema is normative or not.
    
If a schema is not normative, the `XMLParser` uses an informative-only `ErrorHandler`, that reports warnings as `RSC-024` (USAGE), and errors as `RSC-025` (USAGE).
Fatal errors are still reported as `RSC-016` (FATAL).

SVG schema-based content conformance checking is made informative only:
      * new NVDL and RelaxNG schema drivers will only check SVG content, and ignore other
        namespaces.
      * these informative schemas are enabled for both SVG and XHTML Content Documents
      * SVG content conformance, as before, is using RelaxNG schemas from the Nu HTML Checker.


rdeltour:
That's no longer the case in v5.0.0-beta-3, it will be removed in 2c429d0a (big URL refactoring)

PR #1367, opened 2022-12-01
feat: allow SVG/MathML doctype declarations

### LINKED ISSUE #1114: SVG DTD not accepted
(This seems to be related to https://github.com/w3c/epubcheck/issues/173#issuecomment-27623760, although not to `rdf:RDF`, which is the original topic of that issue.)

The [W3C SVG Logo](https://www.w3.org/StyleSheets/TR/2016/logos/W3C.svg) is not accepted by epubcheck, complaining about "External DTD entities are not allowed.". That is problematic.

The SVG file itself does not include any external DTD entity. However, the Logo file does contain:

```
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
```

Indeed, the [DTD file itself](https://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd) contains loads of entities. It is also correct that the latest version of SVG (SVG2) has removed the reference to a DTD (see [change section](https://www.w3.org/TR/SVG2/changes.html#svgdtd)). But flagging this is as an error does create backward incompatibility issues with bona fide [SVG1.1](https://www.w3.org/TR/SVG11/) content which did rely on the DTD (e.g., _all_ examples in SVG1.1 include that DTD reference, but also all SVG content produced by earlier versions of Adobe Illustrator, to take just this example). Note that all browsers accept such SVG 1.1 content without further ado.

I would think that epubcheck should simply ignore the SVG DTD and shouldn't report any error or warning on that one.

### LINKED ISSUE #1192: Select external identifiers allowed in resources
EPUB 3.3 allows a [reserved set of external identifiers](https://w3c.github.io/epub-specs/epub33/core/#app-identifiers-allowed) to be included in resources now.

External entities are still [forbidden in the internal subset](https://w3c.github.io/epub-specs/epub33/core/#sec-xml-constraints).

See the [pull request](https://github.com/w3c/epub-specs/pull/1368) for this one, as the issue(s) got a bit out of hand.

rdeltour:
EPUB 3.3. now allows a reserved set of external identifiers in doctype declarations of documents with select media types.

See: https://www.w3.org/TR/epub-33/#app-identifiers-allowed

This commit:
- adds those as special cases to the XML parser code
- totally removes entity fetching for EPUB 3.3
- keeps forbidding external entities in the internal subset

Fix #1192, Fix #1114


mattgarrish:
The NCX doctype is also allowed per https://www.w3.org/TR/epub-33/#app-identifiers-allowed

Or are you adding that in a separate PR?

rdeltour:
> Or are you adding that in a separate PR?

yes, that was only introduced in v5.0.0-beta-3, in f3e5f673

PR #1368, opened 2022-12-01
feat: report reserved string in XHTML custom attribute namespaces

### LINKED ISSUE #1190: New custom attributes namespace restrictions
EPUB 3.3 disallows w3.org and idpf.org from [custom attribute namespaces](https://w3c.github.io/epub-specs/epub33/core/#sec-xhtml-custom-attributes). See https://github.com/w3c/epub-specs/issues/1388

This is a change from 3.2 where only the following two specific URIs were forbidden:

- http://www.w3.org/1999/xhtml
- http://www.idpf.org/2007/ops

rdeltour:
EPUB 3.3 disallows "w3.org" and "idpf.org" in domains of XHTML custom attribute namespaces.

Fix #1190


mattgarrish:
This still feels kind of weird as we're banning legitimate W3C attributes, too. I guess it doesn't matter, though, since they're technically invalid "custom" attributes until we allow them in the spec's extension section.

(I don't know what this actually affects, though, as there is a non-prefixed version of its:translate in html, which is the only thing I can think of.)

mattgarrish (approved):

PR #1377, opened 2022-12-01
feat: page-list nav does not have to match reading order

### LINKED ISSUE #1237: Page list ordering does not have to match content
The requirement for it to match the spine and document order of the page breaks has been removed.

See https://github.com/w3c/epub-specs/issues/1500

rdeltour:
This commit:
- removes the call to check the page list reading order
- keeps the checking code, as well as NAV-011
- update tests

Fix #1237

### DIFF
--- a/src/test/resources/epub3/navigation-publication.feature
+++ b/src/test/resources/epub3/navigation-publication.feature
@@ -71,14 +71,12 @@ Feature: EPUB 3 ▸ Navigation Document ▸ Full Publication Checks
     When checking EPUB 'nav-page-list-reading-order-valid'
     Then no errors or warnings are reported
 
-  Scenario: Report a `page-list nav` whose links do not match the spine order 
+  Scenario: Verify a `page-list nav` whose links do not match the spine order 
     When checking EPUB 'nav-page-list-unordered-spine-warning'
-    Then warning NAV-011 is reported
     And no other errors or warnings are reported
 
-  Scenario: Report a `page-list nav` whose links do match the document order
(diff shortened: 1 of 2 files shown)


src/test/resources/epub3/navigation-publication.feature
```diff
@@ -71,14 +71,12 @@ Feature: EPUB 3 ▸ Navigation Document ▸ Full Publication Checks
     When checking EPUB 'nav-page-list-reading-order-valid'
     Then no errors or warnings are reported
 
-  Scenario: Report a `page-list nav` whose links do not match the spine order 
+  Scenario: Verify a `page-list nav` whose links do not match the spine order 
     When checking EPUB 'nav-page-list-unordered-spine-warning'
-    Then warning NAV-011 is reported
     And no other errors or warnings are reported
```
mattgarrish, line 76: I guess this doesn't matter much practically for processing, but shouldn't this line start with "Then no errors..." instead of "And no other errors..."?

(Same would apply to line 80.)

rdeltour: Yeah, all the [steps keywords](https://cucumber.io/docs/gherkin/reference/#keywords) can be used interchangeably, so it’s mostly a matter of taste. I personally find “Then warning X is reported AND no other errors or warnings are reported” reads more naturally than repeating “Then”, but YMMV 😊

rdeltour: oh, no I understand just now. What I have is "When … And …" when it would obviously read better as "When … Then …", you're absolutely right.
Filed as #1386

PR #1391, opened 2022-12-02
fix: allow HTML `script` element in EPUB 2.0.1

rdeltour:
EPUB 2.0.1 does not strictly forbids scripts in HTML. It does say that a reading system should not execute the script, but there is no conformance statement on the content side.

This was reported as `SCP-004`, now suppressed.

Note: this change will only be effective when the checks of the `ctc` package  are disabled.


mattgarrish:
Ya, it was experimental, not disallowed, in 2.0.1 to have scripting.

mattgarrish (approved):

PR #1394, opened 2022-12-02
feat: re-implement relevant checks from the `ctc` package in core handlers

rdeltour:
## reimplement HTM-046 and HTM-047 in OPSHandler30

`HTM-046` and `HTM-048` check that HTML fixed-layout documents
have viewport metadata.

This change reimplements these checks in `OPSHandler30` so that the
`ctc` package can be disabled.

## report `RSC-018` as `RSC-007` instead

RSC-018 was introduced in the `ctc` package to check the images
specified in the MathML `altimg` attribute exist in the container.

EPUBCheck already has an error code  for missing resources, namely `RSC-018`.

This  change adds a check for the `altimg` in OPSHandler, and suppresses
`RSC-018`.

## move HTM-053 check to OPSHandler

`HTM-053` is an INFO-level message that reports links to local files
(absolute `file` URLs).

This check is moved to the core OPSHandler checks, so we can
safely disable the `ctc` package.

Note: these changes will only be effective when the `ctc` package is removed.

### DIFF
--- a/src/main/java/com/adobe/epubcheck/ops/OPSHandler30.java
+++ b/src/main/java/com/adobe/epubcheck/ops/OPSHandler30.java
@@ -325,6 +325,11 @@ public class OPSHandler30 extends OPSHandler
       requiredProperties.add(ITEM_PROPERTIES.MATHML);
       inMathML = true;
       hasAltorAnnotation = (null != e.getAttribute("alttext"));
+      String altimg = e.getAttribute("altimg");
+      if (altimg != null) {
+        super.checkImage(e, null, "altimg");
+      }
+        
     }
     else if (name.equals("svg"))
     {
--- a/src/main/java/com/adobe/epubcheck/ops/OPSHandler30.java
+++ b/src/main/java/com/adobe/epubcheck/ops/OPSHandler30.java
@@ -3,6 +3,7 @@ package com.adobe.epubcheck.ops;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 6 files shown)


src/main/java/com/adobe/epubcheck/ops/OPSHandler30.java
```diff
@@ -679,6 +692,31 @@ else if (!isOutermostSVGAlreadyProcessed)
...
+      if ("viewport".equals(name))
+      {
+        hasViewport = true;
+        // For a fixed-layout document:
+        // check that the vieport declares both height and width
+        if (context.opfItem.isPresent() && context.opfItem.get().isFixedLayout())
+        {
+          String contentAttribute = e.getAttribute("content");
+          if (contentAttribute == null || !contentAttribute.contains("width=")
+              || !contentAttribute.contains("height="))
```
mattgarrish, line 709: It looks like this may fail some valid cases, like " height = 1200, width = 800".

The syntax allows spaces on either side of the equals sign: https://www.w3.org/TR/epub-33/#viewport.ebnf.assign

mattgarrish (changes requested):

src/main/java/com/adobe/epubcheck/ops/OPSHandler30.java
rdeltour: Yeah, this PR basically ported the old checks as-is, good catch!
More accurate parsing of the `viewport` syntax is implemented in #1400, with newer changes not being fully taken into account but tracked as #1401.

mattgarrish (approved):

PR #1396, opened 2022-12-02
feat: properly implement URL parsing

rdeltour:
This PR significantly refactors the XML parser/handler code, as well as implements proper URL parsing throughout EPUBCheck's code.

## URL parsing

EPUBCheck now uses URL records internally for everything related
to paths and URLs handling.

We use the Galimatias library which is very good, close to the URL
standard, and also used in Validator.nu.

This change also include many other improvements to the code and
internal APIs used in EPUBCheck checkers.

BREAKING CHANGE: significant refactoring of the internal URL handling

## Refactoring

- remove XMLParser#getResourceName() method
- remove unused VersionRetriever class
- replace publication type strings with enum
- cleanup XMLParser (and handlers) code
  - extract distinct responsibilities into dedicated classes
  - the XMLParser uses a single DelegateHandler class that delegates
to registered handlers
  - handlers no longer use the XMLParser to access the parsing context
(XML model and current location), it all comes from inherited handler
implementations
  - the XMLHandler inteface is now an abstract class that is responsible
for building the EPUBCheck XML model (XMLElement and XMLAttribute)
  - reorganize the package: model and handlers are in dedicated subpackages
  - use more modern try-with-resource pattern to close input streams
- extract preprocessing code to HTMLUtils
- make OPFPeeker use XMLParser: We used to maintain two very similar parsing logic configurations, one for checkers XML parsers and one for the peekers (OPFPeeker). This change unifies them by making XMLParser usable as an XML peeker (i.e. a silent parser which can abort early).
  - XMLParser now has a `#setReporting(boolean reporting)` method which
    can be used to disable any message reporting.
  - a ContentHandler can throw a new SAXAbortException exception to abort
    the parsing silently (without reporting a message).
  - an aborting handler is responsible to report any error or situation
    that caused the parser to abort. For instance, in OPFPeeker, the
    abort cause is stored in the data builder, which will throw exceptions
    when built.
- maintain base URL in a dedicated handler
- simplify profile/type matching check


rdeltour:
@mattgarrish I don't expect a full code review here 😅 I added you as a reviewer mostly so that you're aware of these changes.

mattgarrish:
Ya, I was just going to try my old fallback of running it against the samples to see if anything weird popped up for this one! 😉

rdeltour:
> I got an error building this one from the epub2/ncx-publication.feature file, which doesn't appear in the list of changed files

yeah, not quite sure what happens here. I usually run the whole test suite before making a "feature" commit, but apparently missed this one. 

it definitely passed or was updated for v5.0.0-beta-2 when this change was introduced, so probably a temporary mishap here.

mattgarrish (approved):
I opened a couple of issues that I found in the test samples (cfis and fragment references), as not sure if you want to address them here.

PR #1419, opened 2022-12-02
fix: consider HTML form elements as scripted content

### LINKED ISSUE #1282: html forms are not checked when determining if a file is scripted
I understand that scripted document are ones containing script elements "and/or HTML forms", according to [1]. However, epubcheck seems to ignore the `<form>` element when checking the correct use of `scripted`.

~On the other hand, the rest of the spec seems to use the word "scripted" only for actual scripts, not forms, so the issue is not entirely clear to me. Could it be a spec problem?~ The 3.3 draft is clearer about this [2].

[1] https://www.w3.org/publishing/epub3/epub-packages.html#scripted
[2] https://www.w3.org/TR/epub-33/#sec-scripted-content

rdeltour:
Fixes #1282


mattgarrish:
Scratch what I said above. After digging around through some old issues this is fine. I've opened https://github.com/w3c/epub-specs/issues/2497 to clarify the ambiguity in the spec, though.

PR #1427, opened 2022-12-05
feat: no longer check collection `role`  

### LINKED ISSUE #1350: No restrictions on collection roles
The requirement not to have idpf.org in a collection's role value has been removed. There's also no longer a restriction to using only the values defined in the old idpf registry. You can use either a nmtoken or a url in the attribute.

I've removed the tests that were checking the above two cases, but the code checking them also needs removing.

rdeltour:
This PR suppresses and removes the code of the following checks:
  - `OPF-068`: checked that a role token were one the defined roles in
    the collection role registry
    (see https://idpf.github.io/epub-registries/roles/)
  - `OPF-069`: checked that a custom role URL did not contain the string
    "idpf.org".
  
  These restrictions were removed in EPUB 3.3.
  
  The attribute value is still tokenized, and any URL-looking string
    (in the shape `scheme:something`) is checked to be a valid URL string.
  
  Fix #1350.


mattgarrish (approved):
The comments in the test file are a bit misleading - nothing has changed about how collection roles are validated between 3.2 and 3.3. You don't have to maintain the 3.2 rules anymore as no one uses these anyway, but it's more that collections are being abandoned without deprecating than anything about them has changed.

rdeltour:
Right. I clarified the comments in #1447

PR #1433, opened 2022-12-05
fix: improve checking of missing SVG link labels, now USAGE

### LINKED ISSUE #1353: warning for "SVG hyperlinks should have a human-readable title" seems too strict
Using v5.0.0-beta-2, the following minimal example produces 2 warnings of the following form:

> WARNING(ACC-011): [..] SVG hyperlinks should have a human-readable title (using the "title" child element or the "xlink:title" attribute).

This seems overly strict - aria-label and text descendants are valid ways to label a link in SVG.

```html
<?xml version="1.0" encoding="utf-8"?>
<!DOCTYPE html>

<html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops">
<head>
  <title>SVG anchor test</title>
</head>

<body>
<svg viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg">
  <a href="https://developer.mozilla.org/en-US/docs/Web/SVG/Element/circle" aria-label="circle">
    <circle cx="50" cy="40" r="35" />
  </a>
  <a href="https://developer.mozilla.org/en-US/docs/Web/SVG/Element/text">
    <text x="50" y="90" text-anchor="middle">text</text>
  </a>
</svg>

</body>
</html>
```

rdeltour:
This commit changes the severity of error `ACC-011` to USAGE (was WARNING). Labelling links is not strictly a spec requirement, and checking for missing labels should be done with specialized accessibility inspection.

The check is still available, with an improved logic: the link is reported as having a missing accessible label only if:
- it does not have an `xlink:title` attribute
- AND it does not have an `aria-label` attribute
- AND it does not have a `title` element child
- AND it does not have a `text` element child

Note that his logic is incomplete, as it does not check that the content of `title` or `text` children is not empty. But reporting links that have none of the above can be somewhat useful as a simplified first-pass accessibility check.

Fix #1353


mattgarrish (approved):
Checking that there is at least some text only begins to address the accessibility issue, which is about being able to understand the link's purpose. Fine with this being a usage message as it belongs in ace to be further checked when doing a full evaluation.

PR #1435, opened 2022-12-05
feat: remove checks related to deprecated bindings feature

rdeltour:
The use of the `bindings` element was already reported as a warning (it is deprecated since EPUB 3.2).
  Since it is not necessary to report errors for misuse of a long-deprecated feature, this PR removes the following checks:
  - `OPF-008` (was reported for handler binding to a core media type)
  - `OPF-009` (was reported when a media type was assigned multilpe handlers)
  - `OPF-046` (was reported when a media type hander was not declared as scripted)
  - a schema warning (`RSC-017`) reported when a handler was not XHTML

See also issue #881


mattgarrish (approved):
bindings and triggers were both for a kind of reading system that never came into existence, so this is fine by me.

PR #1464, opened 2022-12-15
feat: check that non-linear content is reachable

### LINKED ISSUE #1451: check that non-linear content is reachable
The spec says:

> EPUB creators MUST provide a means of accessing all non-linear content (e.g., hyperlinks in the content or from the [EPUB navigation document](https://www.w3.org/TR/epub-33/#sec-nav)).EPUB creators MUST provide a means of accessing all non-linear content (e.g., hyperlinks in the content or from the [EPUB navigation document](https://www.w3.org/TR/epub-33/#sec-nav)).EPUB creators MUST provide a means of accessing all non-linear content (e.g., hyperlinks in the content or from the [EPUB navigation document](https://www.w3.org/TR/epub-33/#sec-nav)).

That does not seem to be checked by EPUBCheck.

rdeltour:
EPUB 3.3 says:
> EPUB creators MUST provide a means of accessing all non-linear content
> (e.g., hyperlinks in the content or from the EPUB navigation document).

This commit adds a check for the above statement.
- new error `OPF-096` is reported when no hyperlink was found to non-linear content in an EPUB with no script
- new usage `OPF-096b` is reported when no hyperlink was found to non-linear content in a scripted EPUB (a link may be added via scripting, so we cannot report this as an error)

Fix #1451


mattgarrish (approved):
Forgot about this. It was introduced back in 3.1 during one of the arguments over whether reading systems have to show or skip non-linear content.

This seems fine as far as checking obvious links go, but I wonder if it will come back to bite us when scripting comes into play (e.g., if a non-linear document is only reachable in certain scenarios, perhaps like a dynamic choose-your-own-adventure book). Maybe that's so niche it's not worth bothering about, though.

rdeltour:
> This seems fine as far as checking obvious links go, but I wonder if it will come back to bite us when scripting comes into play (e.g., if a non-linear document is only reachable in certain scenarios, perhaps like a dynamic choose-your-own-adventure book). Maybe that's so niche it's not worth bothering about, though.

EPUBCheck will only report a usage in that case (if scripting is found in the EPUB).

A question however: should the Nav Doc be exempt from that check, given that it's always rendered by RS?
I had to fix a few test cases where the Nav Doc was in the spine as non-linear content, but only realize now that it may be just fine.

Also, EPUBCheck will report an error if non-linear content is unreachable even if it is used as fallback content for a reachable document. But I think it's niche enough to not be bothering about that one (until it is reported as a false-positive 😉).

mattgarrish:
> A question however: should the Nav Doc be exempt from that check, given that it's always rendered by RS?

Ya, that's a funny case. I don't know that the reading system rendering of the document would satisfy the requirement, since it's selective in what it will show and what it's capable of rendering. If you put the nav doc in the spine and mark it as non-linear then you want users to get the full html rendering, so I'd expect there should also be a means of accessing the document.

rdeltour:
> If you put the nav doc in the spine and mark it as non-linear then you want users to get the full html rendering, so I'd expect there should also be a means of accessing the document.

Right, makes sense. I'll leave the check as-is then!

Issue #58, opened 2013-10-15
Checking for unused files which are not in the manifest


rdeltour:
_From [dau...@gmail.com](https://code.google.com/u/107924573892939151290/) on March 12, 2010 20:13:51_

This is an extension of issue #38 . ePubCheck does not currently detect
files that are in the zip package but not the manifest, if these files are
not referenced from anywhere in the ePub.

A major new player in the eBook marketplace is now checking this, and
rejecting titles with "unmanifested" files. I know of cases where full
print PDFs accidentally ended up in the ePub zip. This doesn't really cause
any problems, but it does increase file size.

Dave Cramer
Hachette Book Group

_Original issue: http://code.google.com/p/epubcheck/issues/detail?id=58_

PR #1465, opened 2022-12-22
clarify manifest resources checks

### LINKED ISSUE #1452: check that the manifest only lists publication resources
The spec says:

> the manifest MUST only list publication resources

I don't think this is covered by EPUBCheck.

For each manifest item, we could lookup the reference registry to see if a reference to the resource was found, and issue an error if not.
We would issue a mere usage message if the EPUB contains scripted content, as the item could potentially be a legit publication resource used in scripting.

rdeltour:
This PR tries to clarify how we check resources listed in the manifest.

Basically:
- resources in the container but not listed in the manifest are now only reported as a usage (`OPF-003`), instead of a warning
- resources in the `manifest` but for which no reference were found are reported as a usage (new `OPF-097`)

## feat: report as a usage when no reference were found to manifest items  
  
  This commit introduces a new usage message (`OPF-097`) reported when the
  manifest includes a resource for which no reference was found in content.
  
  This is legit for instance for exempt resources (like data files included
  in the container). But it is likely an author error otherwise, so reporting
  a usage message can help to detect that.
  
  We do not report an error since:
  - references can be added via scripting, so the check can only be
    informative for scripted content
  - EPUBCheck historically required all container resources to be listed
    in the manifest (not doing so was reported as a warning, `OPF-003`),
    which is conflicting with a requirement to **not** list unsused
    resources
  
  Close #1452

## feat: report as usage container resources not listed in manifest  
  
  EPUBCheck used to report any resource found in the the container but not
  listed in the manifest as a warning (since #58 was fixed, in v1.1). But
  the EPUB specification does not require that.
  
  This commit downgrades the severity of `OPF-003` to a usage report.
  
  See also #1452
  See also w3c/epub-specs#563


rdeltour:
@mattgarrish hang on before reviewing this, the logic is broken.

mattgarrish (approved):

PR #1467, opened 2022-12-23
feat: check that the manifest is not self-referencing

### LINKED ISSUE #1453: check that the manifest must not list the package document itself
The spec says:

> Note that the manifest is not self-referencing: EPUB creators MUST NOT specify an item element that refers to the [package document](https://www.w3.org/TR/epub-33/#dfn-package-document) itself.

Currently, EPUBCheck loops indefinitely and finally throws a StackOverflowError when the package document refers to itself. We should catch this.

rdeltour:
This commit adds a new check, reported as `OPF-099` (error), to verify that the package document `manifest` does not include an `item` element that refers to the package document itself.

This statement was apparently not checked by EPUBCheck previously. Worse, the code was entering an infinite loop, as the package document checker was creating a new checker for itself, recursively.

Fix #1453

### DIFF
--- a/src/main/resources/com/adobe/epubcheck/messages/MessageBundle.properties
+++ b/src/main/resources/com/adobe/epubcheck/messages/MessageBundle.properties
@@ -215,6 +215,7 @@ OPF_096=Non-linear content must be reachable, but found no hyperlink to "%1$s".
 OPF_096b=No hyperlink was found to non-linear document "%1$s", please check that it can be reached from scripted content.
 OPF_097=Resource "%1$s" is listed in the manifest, but no reference to it was found in content documents.
 OPF_098=The "href" attribute must reference resources, not elements in the package document, but found URL "%1$s".
+OPF_099=The manifest must not list the package document.
 
 #Package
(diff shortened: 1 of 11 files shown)


src/main/resources/com/adobe/epubcheck/messages/MessageBundle.properties
```diff
@@ -215,6 +215,7 @@ OPF_096=Non-linear content must be reachable, but found no hyperlink to "%1$s".
 OPF_096b=No hyperlink was found to non-linear document "%1$s", please check that it can be reached from scripted content.
 OPF_097=Resource "%1$s" is listed in the manifest, but no reference to it was found in content documents.
 OPF_098=The "href" attribute must not reference resources via elements in the package document itself, but found URL "%1$s".
+OPF_099=The package document must not include itself in its manifest.
```
mattgarrish, line 218: A minor clarity suggestion:

```suggestion
OPF_099=The manifest must not list the package document . 
```

mattgarrish (approved):

PR #1469, opened 2022-12-23
feat: check "data-*" attributes name restrictions

### LINKED ISSUE #1107: Check data-* name restrictions
I noticed that we check that data-* attributes are allowed in both XHTML and SVG content documents but we don't appear to be enforcing the naming restrictions in the HTML specification -- specifically that the name can't begin with "xml" or contain upper-ascii characters.

https://html.spec.whatwg.org/multipage/dom.html#embedding-custom-non-visible-data-with-the-data-*-attributes

rdeltour:
According to HTML:
> A custom data attribute is an attribute in no namespace whose name
> starts with the string "data-", has at least one character after the
> hyphen, is XML-compatible, and contains no ASCII upper alphas.

This commit reports invalid `data-*` attributes as errors with a new error code, `HTM-061`.

This should be ideally delegated to validator.nu, but we add this check until we fully integrate the HTML checker.

Close #1107

### DIFF
--- a/src/main/resources/com/adobe/epubcheck/messages/MessageBundle.properties
+++ b/src/main/resources/com/adobe/epubcheck/messages/MessageBundle.properties
@@ -64,6 +64,7 @@ HTM_058=HTML documents must be encoded in UTF-8, but UTF-16 was detected.
 HTM_059=Viewport "%1$s" property must not be defined more than once, but found values [%2$s].
 HTM_060a=EPUB reading systems must ignore secondary viewport meta elements in fixed-layout documents; viewport declaration "%1$s" will be ignored. 
 HTM_060b=EPUB reading systems must ignore viewport meta elements in reflowable documents; viewport declaration "%1$s" will be ignored.
+HTM_061="%1$s" is not a valid custom data attribute (it must have at least one character after the hyphen, be XML-compatible, and not contain ASCII uppercase letters).
 
 #media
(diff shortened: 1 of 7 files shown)


src/main/resources/com/adobe/epubcheck/messages/MessageBundle.properties
```diff
@@ -64,6 +64,7 @@ HTM_058=HTML documents must be encoded in UTF-8, but UTF-16 was detected.
 HTM_059=Viewport "%1$s" property must not be defined more than once, but found values [%2$s].
 HTM_060a=EPUB reading systems must ignore secondary viewport meta elements in fixed-layout documents; viewport declaration "%1$s" will be ignored. 
 HTM_060b=EPUB reading systems must ignore viewport meta elements in reflowable documents; viewport declaration "%1$s" will be ignored.
+HTM_061="%1$s" is not a valid custom data attributes (it must have at least one character after the hyphen, be XML-compatible, and contain no ASCII upper alphas).
```
mattgarrish, line 67: s/attributes/attribute/

I also wonder if "upper alphas" will be clear to general users. Might be safer to be more verbose with "uppercase letters".

```suggestion
HTM_061="%1$s" is not a valid custom data attribute (it must have at least one character after the hyphen, be XML-compatible, and not contain ASCII uppercase letters).
```

mattgarrish (approved):

Issue #1478, opened 2023-01-23
Version 5.0: Exception in thread "main" ...


Doktorchen:
I tested the epubcheck 5 on debian linux.
Instead of testing the book I only got this message:

~/ePub/epubcheck/epubcheck-5.0.0.jar *.epub
Exception in thread "main" java.lang.NoClassDefFoundError: io/mola/galimatias/GalimatiasParseException
        at com.adobe.epubcheck.tool.Checker.main(Checker.java:30)
Caused by: java.lang.ClassNotFoundException: io.mola.galimatias.GalimatiasParseException
        at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
        at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
        ... 1 more

rdeltour:
Did you extract the `.jar` file out of its parent directory? It needs to be next to the `lib` directory so that 3d-party dependencies are in the class path…

Doktorchen:
Presumably the main problem is, that you do not provide a zip-archive anymore including all necessary files and directories as for previous versions.
What I downloaded as version 5 has only a size of 1.2 MiB, previous zip-archives have one order of magnitude more content - including a lib directory, licences, changelog, readme.
Something seems to be missing.
Why not to provide such a complete zip-archive as for previous versions - this requires no further diving into details for people just wanting to check EPUBs.

Doktorchen:
I looked intensively on this page, but found no download at all.
I even looked in the source code of the page for 'zip' or 'ZIP' - without a result.
The SVG with 'downloads@5.0.0' is included in an a element pointing to this SVG file again.
Therefore I looked at this Maven Central Repository page and found only this jar file named epubcheck-5.0.0.jar.
Can you provide a precise link to the zip file? ;o)

rdeltour:
Here’s a direct link:
https://github.com/w3c/epubcheck/releases/download/v5.0.0/epubcheck-5.0.0.zip

I checked on different browsers and different devices and could find the link at the usual place (under the "assets" expandable area). Maybe an extension is hiding it in your setup?

Doktorchen:
Thanks, this zip archive works.

But I cannot see any link in assets, but an animated SVG, some buttons in a form.
[...]

PR #1466, opened 2022-12-23
improve checks of package links

rdeltour:
## fix: allow links to resources embedded in content documents  
  
  This commit checks that the linked resource is not in the spine before
  reporting `OPF-067` ("The resource "…" must not be listed both as a
  "link" element in the package metadata and as a manifest item.").
  
  Effectively, this allows package links to publication resources in the
  two cases identified in the EPUB 3.3 specification:
  
  > Resources referenced from the link element are publication resources
  > only when they are:
  > - referenced from the spine; or
  > - included or embedded in an EPUB content document
  
  Fix #1454

## feat: report package link to package document elements  
  
  This commit introduces a new error message (`OPF-098`) to report `href` attributes
  containing a URL to package document elements.
  
  This implements a check for the following specification statement:
  > The URL string MUST NOT reference resources via elements in the package document

Issue #1444, opened 2022-12-05
Links with epub:type require href


mattgarrish:
You can't use epub:type on an inactive link like this:

```<a epub:type="glossref">foo</a>```

If you do, epubcheck will report:

> Error while parsing file: element "a" missing required attribute "href"

I don't know if this occurs often, but it seems like it might be too strict. I'm more worried that people have used it on bad markup patterns like this from the samples: `<a epub:type="keyword">bar</a>`.

PR #1530, opened 2023-07-06
feat: update to latest schemas from the Nu HTML Checker

rdeltour:
---------------------

Add `plaintext-only` value to `contenteditable` attribute

https://github.com/validator/validator/commit/228ea687b

---------------------

add warnings for deprecated ARIA attributes (#1560)

https://github.com/validator/validator/commit/e61fbeb2c

---------------------

disallow aria attributes on datalist (#1557)

https://github.com/validator/validator/commit/b4849c04b

---------------------

allow role=button on area

https://github.com/validator/validator/commit/89ff1fc2b

---------------------

fix: allow -1 for aria-colcount

https://github.com/validator/validator/commit/2466e73c4

---------------------

fix: Allow -1 for aria-rowcount

https://github.com/validator/validator/commit/eac205b6b

---------------------

Add `fetchpriority` attribute

https://github.com/validator/validator/commit/7c9991405

---------------------

Script type importmap (#1478)

https://github.com/validator/validator/commit/4054c56f5

---------------------

Move “manifest” attribute to legacy-attributes list

https://github.com/validator/validator/commit/175e62d3a

---------------------

Pointer Events (#1481)

https://github.com/validator/validator/commit/5eb52a931

---------------------

Report a better error message for obsolete “manifest” attribute

https://github.com/validator/validator/commit/ae924b625

---------------------

onscrollend

https://github.com/validator/validator/commit/87287a97f

---------------------

slot element

https://github.com/validator/validator/commit/d6cf423c1

---------------------

Allow paragraphs in hgroup (#1467)

https://github.com/validator/validator/commit/3ff53973f

---------------------

blocking="render" (#1466)

https://github.com/validator/validator/commit/df5d4e23a

---------------------

inert

https://github.com/validator/validator/commit/507f551b2

---------------------

hidden=until-found

https://github.com/validator/validator/commit/a424d55a8

---------------------

media attribute in meta element

https://github.com/validator/validator/commit/af3e94d22

---------------------

Sync event handlers

https://github.com/validator/validator/commit/58bea51c7

---------------------

Remove manifest attribute

https://github.com/validator/validator/commit/66510f3ca

---------------------

enterkeyhint

https://github.com/validator/validator/commit/70f3044d7

---------------------

Form method dialog (#1443)

https://github.com/validator/validator/commit/0db229a44

---------------------

imagesrcset and imagesizes in link (#1444)

https://github.com/validator/validator/commit/49656dbfc

---------------------

feat: add support width and height of source in picture

https://github.com/validator/validator/commit/176d98e4f

---------------------

Update: allow none/presentation on nav element (#1323)

https://github.com/validator/validator/commit/5e25edb39

---------------------

Update schema/html5/web-forms.rnc

https://github.com/validator/validator/commit/95ba6c7cf

---------------------

update role allowances for button and input type=button

https://github.com/validator/validator/commit/6cc14c4d7

---------------------

adds role=radio alowance to img alt=foo

https://github.com/validator/validator/commit/b013aa02b

---------------------

remove deprecated dpub role allowances

https://github.com/validator/validator/commit/b0fb152c0

---------------------

allow section element to have role=group

https://github.com/validator/validator/commit/078341d37


rdeltour:
@mattgarrish, I'd like your input on this one.

This new round of update includes https://github.com/validator/validator/pull/1262/commits/cf03b5a4e94df8413a9a27b8a244bbe8c9c06d75, which removed `doc-endnote` and `doc-biblioentry` as allowed roles on `li` (apparently I didn't in fact update the schema since earlier, my bad).

While I understand the reason for the HTML checker, raising an error sound a bit drastic for EPUBCheck. I'm considering re-adding the roles to our version of the schema and raising a bespoke warning instead. WDYT?

mattgarrish:
> raising an error sound a bit drastic for EPUBCheck

Ya, it's not like using the roles really breaks anything. The inheritance model they break is a technical spec issue not a real problem that affects AT.

They shouldn't be used anymore, but 1.1 isn't even a REC yet. A warning sounds better than an error.

mattgarrish:
The test to catch the roles is a bit limited. It only checks for the roles on `li` instead of all elements that accept any role. Because you can have `<div role="list">`, for example, you can have `<div role="doc-endnote">` as a child of it. (See [example 18](https://www.w3.org/TR/dpub-aria-1.0/#doc-endnote) in the 1.0 spec.)

rdeltour:
Yeah I know. That was more or less intentional, as I think this is not used a lot and the HTML validator does not catch these either.
But adding a naive fix (parent/child only) in Schematron should be straightforward.

rdeltour:
> But adding a naive fix (parent/child only) in Schematron should be straightforward.

Or maybe even simpler: shouldn't a warning be raised whenever one of these deprecated role is used, regardless of the semantic of the holding element and ancestor?

mattgarrish:
> shouldn't a warning be raised whenever one of these deprecated role is used, regardless of the semantic of the holding element and ancestor?

Yes, I wouldn't worry about whether there's a list ancestor or not, or what roles the element should accept. That's kind of moot. If any element has a role attribute with doc-endnote or doc-biblioentry, it should raise a warning.

rdeltour:
Done.

mattgarrish (approved):

PR #1543, opened 2023-09-05
fix: use correct german quotes and fixed some mistakes

Wandmalfarbe:
- Used the correct German quotes `„` and `“` in UTF-8 encoded properties files.
- Used the correct German quotes `\u201e` and `\u201c` in ISO-8859-1 encoded properties files.
- Fixed some other mistakes.
- I started to translate some of the English messages in the `MessageBundle_de.properties` file into German, but soon gave up because there were so many missing translations :)


src/main/resources/com/adobe/epubcheck/messages/MessageBundle_de.properties
```diff
@@ -224,55 +224,55 @@ OPF_099=The manifest must not list the package document.
...
 
 #Resources
-RSC_001=Datei "%1$s" wurde nicht gefunden.
-RSC_002=Erforderliche Datei "META-INF/container.xml" fehlt.
-RSC_003=Es wurde kein "rootfile"-Element mit MIME-Type "application/oebps-package+xml" im EPUB gefunden.
-RSC_004=File "%1$s" is encrypted, its content will not be checked.
+RSC_001=Datei „%1$s“ wurde nicht gefunden.
+RSC_002=Erforderliche Datei „META-INF/container.xml“ fehlt.
+RSC_003=Es wurde kein „rootfile“-Element mit MIME-Type „application/oebps-package+xml“ im EPUB gefunden.
+RSC_004=Datei „x“ ist verschlüsselt; die Inhalte werden nicht überprüft.
```
Wandmalfarbe, line 257: Ah, I missed this. It should be `„%1$s“` instead of `„x“`.

firefly9966 (approved):

rdeltour:
Looking at this old PR…
Usually any change to the translations would go through our Transifex translation service (see the [Translating](https://github.com/w3c/epubcheck/wiki/Translating) doc).

Last time I checked I had trouble updating the transifex-hosted translation from the properties file, but that should be doable. I'll keep this PR open until I try again.

PR #1547, opened 2023-10-10
Zip file is not closed properly

Kamik423:
For checking the contents of an EPUB, a `java.util.zip.ZipFile`-object is created. This opens the file for introspection. The file can be closed in two different ways: By explicitly being closed once epubcheck is done with it (calling the `.close()` method), which is *not* done in this case; or by waiting for the garbage collection to clean up the object which results in a close as well. The latter is what happens in epubcheck. However there is no guarantee when Java runs the garbage collection, it could after hours. If epubcheck is integrated into another long running process, this can block EPUB file for quite a long time – during which it can not be deleted or overwritten.

I have implemented a fix for this, storing the ZipFile in the state object and closing it once all checks have been performed. The ZipFile object is still required to be open for checking all the resources during the process so it cannot just be closed at the end of the `checkContainerStructure(···)` method.

A merge would be greatly appreciated.

### DIFF
--- a/src/main/java/com/adobe/epubcheck/ocf/OCFChecker.java
+++ b/src/main/java/com/adobe/epubcheck/ocf/OCFChecker.java
@@ -233,6 +233,18 @@ public final class OCFChecker extends AbstractChecker
     //
     checkFileExtension(state);
 
+    // Close zip file to free resource
+    if (state.getZipResources() != null)
+    {
+      try
+      {
+        state.getZipResources().close();
+      }
+      catch (Exception e)
+      {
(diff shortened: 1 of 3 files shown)


firefly9966 (approved):

rdeltour (approved):
Thanks for the PR @Kamik423, and sorry for the late review.

I'll do a little cleanup when merging with little tweaks (mostly for cosmetics, but also to call the close() method in a finally block). But otherwise looks good to me!

rdeltour:
Rebased and applied as 96a6e1cef9df4a7e70ae0dd80eaf5f7783fcdd16

PR #1576, opened 2024-11-28
fix: recognize `--failonwarnings+` and `--failonwarnings-`

### LINKED ISSUE #1575: epubcheck 5.1.0 does not recognize `--failonwarnings+`, nor `--failonwarnings-`
According to [the help message](https://github.com/w3c/epubcheck/blob/v5.1.0/src/main/resources/com/adobe/epubcheck/util/messages.properties#L90-L92), both `--failonwarnings+` and `--failonwarnings-` should be valid.
But epubcheck 5.1.0 says about them, respectively:
```
Unrecognized argument: "--failonwarnings+"
EPUBCheck v5.1.0
...
```
and
```
Unrecognized argument: "--failonwarnings-"
EPUBCheck v5.1.0
...
```
respectively.

naglis:
Fixes #1575

Although I don't understand the purpose of the extra `+` or `-` - i.e. does it not simply repeat what specifying `--failonwarnings` and omitting `--failonwarnings` would achieve?

### DIFF
--- a/src/main/java/com/adobe/epubcheck/tool/EpubChecker.java
+++ b/src/main/java/com/adobe/epubcheck/tool/EpubChecker.java
@@ -663,6 +663,8 @@ public class EpubChecker
               outWriter.setQuiet(true);
             break;
           case "failonwarnings":
+          case "failonwarnings+":
+          case "failonwarnings-":
               String fw = args[i].substring("--failonwarnings".length());
               failOnWarnings = (fw.compareTo("-") != 0);
             break;


rdeltour:
Thanks for the PR @naglis! 

Yeah, I'm not sure for how long we've had this option and it's been documented this way, but I agree that having both `--failonwarnings+` and `--failonwarnings-` looks a bit odd.

I'll close this PR and will create another one to keep only a `--failonwarnings` option (which matches the current CLI anyways).

Issue #1554, opened 2024-02-05
java.lang.ArrayIndexOutOfBoundsException


Matt-1:
I have an EPUB for which epubcheck v5.1.0 runs into an error:
```
Validating using EPUB version 3.3 rules.
java.lang.ArrayIndexOutOfBoundsException: 2048
        at org.apache.xerces.impl.io.UTF8Reader.read(Unknown Source)
        at org.apache.xerces.impl.XMLEntityScanner.load(Unknown Source)
        at org.apache.xerces.impl.XMLEntityScanner.scanContent(Unknown Source)
        at org.apache.xerces.impl.XMLDocumentFragmentScannerImpl.scanContent(Unknown Source)
        at org.apache.xerces.impl.XMLDocumentFragmentScannerImpl$FragmentContentDispatcher.dispatch(Unknown Source)
        at org.apache.xerces.impl.XMLDocumentFragmentScannerImpl.scanDocument(Unknown Source)
        at org.apache.xerces.parsers.XML11Configuration.parse(Unknown Source)
        at org.apache.xerces.parsers.XML11Configuration.parse(Unknown Source)
        at org.apache.xerces.parsers.XMLParser.parse(Unknown Source)
        at org.apache.xerces.parsers.AbstractSAXParser.parse(Unknown Source)
        at org.apache.xerces.jaxp.SAXParserImpl$JAXPSAXParser.parse(Unknown Source)
        at org.apache.xerces.jaxp.SAXParserImpl.parse(Unknown Source)
        at com.adobe.epubcheck.xml.XMLParser.process(XMLParser.java:176)
        at com.adobe.epubcheck.ops.OPSChecker.validate(OPSChecker.java:127)
        at com.adobe.epubcheck.ops.OPSChecker.checkContent(OPSChecker.java:97)
        at com.adobe.epubcheck.opf.PublicationResourceChecker.check(PublicationResourceChecker.java:45)
        at com.adobe.epubcheck.opf.OPFChecker.checkItemContent(OPFChecker.java:398)
        at com.adobe.epubcheck.opf.OPFChecker.checkPackage(OPFChecker.java:136)
        at com.adobe.epubcheck.opf.OPFChecker30.checkPackage(OPFChecker30.java:67)
        at com.adobe.epubcheck.opf.OPFChecker.check(OPFChecker.java:94)
        at com.adobe.epubcheck.ocf.OCFChecker.check(OCFChecker.java:173)
[...]

PR #1579, opened 2024-12-23
fix: rare decoding error on UTF-8 documents

### LINKED ISSUE #1548: Unjustified error message "Invalid byte 2 of 4-byte UTF-8 sequence"
I've encountered this error and created a stripped-down version of the original EPUB that still exhibits the error: [invalidUtf8Sequence.epub](http://devel.springer.de/invalidUtf8Sequence.epub)

For this file, EPUBCheck v5.1.0 reports

```
Validating using EPUB version 3.3 rules.
FATAL(RSC-016): invalidUtf8Sequence.epub/OEBPS/html/Chapter_6.xhtml(88,792): Fatal Error while parsing file: Invalid byte 2 of 4-byte UTF-8 sequence.
[...]
```
I believe this is a false positive. At least I can't find anything wrong with the HTML file.

rdeltour:
On rare occasions, decoding UTF-8 documents caused a fatal error RSC-016 (`Invalid byte 2 of 4-byte UTF-8 sequence.`).

This was likely due to a bug in the Xerces XML parser decoding component, see https://issues.apache.org/jira/browse/XERCESJ-1668

As a workaround, we now read documents using the Java built-in UTF-8 decoder instead of Xerces's own decoder, by creating the SAX parsers' InputSource from an InputStreamReader instead of the raw InputStream.

Fixes #1548

### DIFF
--- a/src/main/java/com/adobe/epubcheck/xml/XMLEncodingSniffer.java
+++ b/src/main/java/com/adobe/epubcheck/xml/XMLEncodingSniffer.java
@@ -108,8 +108,28 @@ public final class XMLEncodingSniffer
     return encoding.toUpperCase(Locale.ROOT);
   }
 
+  /**
+   * Checks if the parameter input stream has a UTF-8 byte order mark.
+   *
+   * @param in
+   *        an input stream
+   * @return <code>true</code> if and only if the input stream starts with a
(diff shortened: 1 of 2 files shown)


rdeltour:
Also fixes #1554

PR #1614, opened 2025-05-30
Added application/x-font-ttf to the list of allowed media types

iherman:
These are the possible changes following https://github.com/w3c/epub-specs/issues/667 and https://github.com/w3c/epub-specs/pull/2726.

I followed the advice of @mattgarrish for the files to be changed (https://github.com/w3c/epub-tests/pull/307#issuecomment-2916256569 and https://github.com/w3c/epub-tests/pull/307#issuecomment-2919264506). Note, however, that I was not sure what is expected in the two OPF files, mostly what the value of the `href` attribute should be. I probably got it wrong, it certainly needs a review... But I hope that, nevertheless, this PR helps...

Note that this change is related to EPUB 3.4. I was looking for a label for 3.4, but have not found any (and I did not want to create one, I am just a guest...). I hope I got the other labels right.


mattgarrish:
> We need a different media type and I assume not a completely fake one.

The best I can find is to switch it to `application/x-font-woff` and replace the font file. That seems to have been a valid woff media type at one time.

> Is `mvn` a java thing?

It's the [Apache Maven](https://maven.apache.org/index.html) build tool. I don't think it's exclusively for java, but that's how epubcheck is designed to be compiled.

mattgarrish:
There are some brief instructions on building epubcheck in the [repository's readme](https://github.com/w3c/epubcheck?tab=readme-ov-file#building-epubcheck).

PR #1660, opened 2026-05-01
Update the Dockerfile build script & README.md

novalagung:
The current `Dockerfile` contains outdated docker image that would produces error during docker image build process. This PR contains several changes:

- Fix the deprecated openjdk docker image, replace it with eclipse-temurin.
- Use `printf` rather than `echo` to fix error `The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8)` when running epubcheck via docker container.
- Update docker instruction on README.md file for better clarity

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,17 @@
 # build the epubcheck.jar file
-FROM maven:slim AS builder
+FROM maven:3-eclipse-temurin-21-alpine AS builder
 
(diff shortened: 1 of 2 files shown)


rdeltour:
I updated the Java images to the latest LTS (Java 25), rebased and merged 🚀. Many thanks for the PR!

PR #1681, opened 2026-08-21
Fix typo in OPF-074 error message

acabal:


rdeltour:
Merged. Thanks for the PR!

PR #1720, opened 2026-09-14
fix(CLI): do not print final message after stack overflow

### LINKED ISSUE #1673: Stack overflow with the Oracle runtime environment
I'm on a Windows machine running Java(TM) SE Runtime Environment (build 1.8.0_471-b09) from Oracle.

I run EPUBCheck, messages fly past and then I see the happy result:

`
Messages: 0 fatals / 0 errors / 0 warnings / 0 infos
`
However, if I think to scroll all the way back to the top of the messages I see that I had a stack overflow:

```
>java -jar epubcheck.jar 9780472901234.epub
Validating using EPUB version 3.3 rules.
java.lang.StackOverflowError
        at com.thaiopensource.relaxng.pattern.InterleavePattern.checkRestrictions(InterleavePattern.java:35)
        at com.thaiopensource.relaxng.pattern.InterleavePattern.checkRestrictions(InterleavePattern.java:35)
```

So I run EPUBCheck with `-Xss` to increase the thread stack size:

`java -Xss10m -jar epubcheck.jar 9780472901234.epub`

and the result is:

```
Check finished with errors
Messages: 1 fatal / 238 errors / 562 warnings / 0 infos

```

I guess three parts to this issue:

- There was a stack overflow using the 32-bit JRE from java.com (the default offering)
- I did not know there was a stack overflow until I happened to scroll all the way to the top.
- What I actually saw was that EPUBCheck reported that there were 0 issues.

rdeltour:
Fix the CLI to not try and catch any `Throwable`, only `Exceptions`.

Also, move the printing of the final result outside the finally block and just before returning the value, to make sure it is not printed if an unexpeceted JVM error is thrown.

See #1673

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


rdeltour:
Close #1673

PR #1722, opened 2026-09-15
feat: add schema validation of ITS attributes

### LINKED ISSUE #1616: Allow Internationalization Tag Set attribute in EPUB
The [International Tag Set](https://www.w3.org/TR/its20/) specification defines a number of attributes that are officially usable [in HTML](https://www.w3.org/TR/its20/#html5-local-attributes). (Essentially, the attributes are of the form `its-*`.) As they are valid in HTML, they should be valid in EPUB as well (without using namespaces for that purpose).

(Context: I am reviewing the code of my respec->EPUB3 converter tool, and I tested it on the [clreq note](https://www.w3.org/TR/clreq/). The generated epub displays properly on, say, Thorium, but epubcheck rejects it because it heavily uses the `its-locale-filter-list` attribute. I believe it should not.)

@rdeltour @mattgarrish

rdeltour:
Fixes #1616


src/main/resources/com/adobe/epubcheck/schema/30/mod/its2/its20-html5.rnc
```diff
@@ -0,0 +1,143 @@
...
+
+its-attribute.term.html = attribute its-term { its-term.type }
+
+its-attribute.termConfidence.html = attribute its-term-confidence { its-termConfidence.type }
+
+its-attribute.withinText.html = attribute its-within-text { its-withinText.type }
+
+its-attribute.domainMapping.html = attribute its-domain-mapping { its-domainMapping.type }
+
+its-attribute.taConfidence.html = attribute its-ta-confidence { its-taConfidence.type }
```
utafrali, line 40: `its-attribute.domainMapping.html` is defined here but is never referenced anywhere in the `its-html-attributes` pattern below. Because RELAX NG only permits attributes explicitly listed in the interleave pattern, any document using the valid ITS 2.0 attribute `its-domain-mapping` will be rejected with a false "attribute not allowed" error. Add `its-attribute.domainMapping.html?` to the `its-html-attributes` pattern (likely alongside `its-attribute.withinText.html?`), or if EPUB 3.4 intentionally excludes domain mapping, add a comment explaining why.

src/main/resources/com/adobe/epubcheck/schema/30/mod/its2/its20-html5-types.rnc
```diff
@@ -0,0 +1,151 @@
...
+
+its-revTool.type = string
+
+its-revToolRef.type = datatype.IRI
+
+its-provRef.type = list { datatype.IRI+ }
+
+its-provenanceRecordsRef.type = datatype.IRI
+
+its-idValue.type = string
```
utafrali, line 108: `its-idValue.type` (line 108) and `its-space.type` (line 110) are defined here but have no corresponding attribute definitions in `its20-html5.rnc` and cannot be reached from the `its-html-attributes` pattern. If their omission is tied to the FIXME about stand-off markup in `<script>`, a short inline comment to that effect would prevent future contributors from wondering whether these were accidentally dropped. If they are simply unused, removing them avoids dead schema code.

src/main/resources/com/adobe/epubcheck/schema/30/mod/its2/its20-html5-types.rnc
```diff
@@ -0,0 +1,151 @@
...
+  "inconsistency" | "grammar" | "legal" | "register" |
+  "locale-specific-content" | "locale-violation" | "style" |
+  "characters" | "misspelling" | "typographical" | "formatting" |
+  "inconsistent-entities" | "numbers" | "markup" | "pattern-problem" |
+  "whitespace" | "internationalization" | "length" | "non-conformance" | "uncategorized" |
+  "other"
+
+its-locQualityIssueComment.type = string
+
+its-locQualityIssueSeverity.type =  xsd:double { minInclusive = "0" maxInclusive = "100" }
```
utafrali, line 125: Two whitespace issues on this line: there is a double space after `=` (`its-locQualityIssueSeverity.type =  xsd:double`), and a trailing space after the closing `}`. The same trailing space exists on line 131 (`its-locQualityRatingScore.type`). Neither affects schema semantics, but they are inconsistent with the rest of the file and will trip any future whitespace lint check.

src/test/resources/epub3/07-content-documents/content-documents-xhtml.feature
```diff
@@ -778,6 +778,14 @@ Feature: EPUB 3 — Content Documents — XHTML
     When checking document 'attrs-its-valid.xhtml'
     Then no errors or warnings are reported
 
+  @spec @xref:sec-xhtml-its
+  Scenario: Verify ITS attributes are checked for conformance
+    When checking document 'attrs-its-schema-error.xhtml'
+    Then the following errors are reported:
```
utafrali, line 784: The error scenario covers an unknown attribute and an invalid enum value, which is a solid baseline. However, there is no test confirming that `its-domain-mapping` is accepted (or deliberately rejected). Once the pattern omission noted above is resolved, adding a positive test case for `its-domain-mapping` in `attrs-its-valid.xhtml` would lock in that the attribute does not regress to a false positive.

utafrali (commented):
The architectural change is correct: replacing the silent `its-*` attribute filter with proper schema validation is the right direction, and the RELAX NG schema files are a faithful adaptation of the ITS 2.0 HTML5 rules. The main concrete problem is that `its-attribute.domainMapping.html` is defined in `its20-html5.rnc` but missing from the `its-html-attributes` interleave pattern, meaning any valid EPUB that uses `its-domain-mapping` will trigger a false validation error.

